PR #10, opened 2018-04-25
Setup nuget packaging

BeeWarloc:
The `pack.sh` script now generates a bunch of nuget packages. One for every dll, for referencing from projects and using as a library, and one called `FuseOpen.Uno.Tool`.

I'm currently in the process of testing these packages with Fuse Studio, for now they're available from a custom feed at https://www.myget.org/F/fuseopen-testing/api/v3/index.json


BeeWarloc:
Closing this as its no longer connected to its fork. Will open another one to replace it.

PR #18, opened 2018-04-27
Dynamically load backends from Uno.Build.Targets

BeeWarloc:
This is mostly a pragmatic change to reduce the number of dependencies from Uno.Build.Targets nuget package.

The list of backends supported is hardcoded with strings, added some tests to ensure this won't accidentally break in the future.


BeeWarloc:
Closing this as its no longer connected to its fork.
Can reopen another PR with this in the future, if making this change at some point becomes a high enough priority.

PR #23, opened 2018-05-03
Unify build across platforms

BeeWarloc:
Here's a set of changes that allows uno built on Windows work on macOS, and vice versa.

- Distribute both Windows and macOS app loaders regardless of build platform
  - Move each apploader to its own subdirectory to avoid filename conflicts
  - Remove touch screen support on Windows, to get rid of `Microsoft.Ink` dependency
- Get rid of `#if UNIX` regions from `Uno.Build.Targets` and instead do a runtime check
- Include prebuilt monostub in repo to avoid depending on native macOS build environment

To not do too much at the same time and break existing CI infrastructure, there's still two solution files. However, I anticipate that we'll be able to get by with just one solution file in the near future.


BeeWarloc:
Hmm.. Seems like the Windows apploader depends on `Microsoft.Ink` which is not distributed with mono. I'll see if that's available as a NuGet package instead.

BeeWarloc:
> Hmm.. Seems like the Windows apploader depends on Microsoft.Ink which is not distributed with mono. I'll see if that's available as a NuGet package instead.

Unfortunately there isn't a NuGet package available for this. That means making Uno.AppLoader-WinForms build on macOS will be a bit of a hassle.

Not sure what I'm going to do next with this yet.

kusma:
I'm not 100% sure the Microsoft.Ink dependency is essential. I think it's for touch-screen support on Windows, which isn't really a supported feature any more. It was used for some customer project at some point, I think.

BeeWarloc:
Ok, I've removed the Windows touch support for now. It's probably not used much, if at all, and it won't affect touch support in Fuse Studio (which I don't know if we have or not).

BeeWarloc:
There's another problem when running the apploaders from within the solution, making this red. Working on a fix for that now.

kusma:
```
Library\Core\UnoCore\Targets\DotNet\Executable.uxl(26.14): E0000: Could not find file 'C:\projects\uno\src\runtime\Uno.AppLoader-WinForms\bin\Debug\apploader-win\Uno.Support.OpenTK.dll'.
Library\Core\UnoCore\Targets\DotNet\Executable.uxl(27.14): E0000: Could not find file 'C:\projects\uno\src\runtime\Uno.AppLoader-WinForms\bin\Debug\apploader-win\OpenTK.dll'.
Library\Core\UnoCore\Targets\DotNet\Executable.uxl(31.14): E0000: Could not find file 'C:\projects\uno\src\runtime\Uno.AppLoader-WinForms\bin\Debug\apploader-win\x64\libEGL.dll'.
Library\Core\UnoCore\Targets\DotNet\Executable.uxl(32.14): E0000: Could not find file 'C:\projects\uno\src\runtime\Uno.AppLoader-WinForms\bin\Debug\apploader-win\x64\libGLESv2.dll'.
Library\Core\UnoCore\Targets\DotNet\Executable.uxl(33.14): E0000: Could not find file 'C:\projects\uno\src\runtime\Uno.AppLoader-WinForms\bin\Debug\apploader-win\x64\d3dcompiler_47.dll'.
```
I guess `Library\Core\UnoCore\Targets\DotNet\Executable.uxl` needs some update as well?

BeeWarloc:
Still broken 😢
Getting the same issue locally too, though. So should be able to figure out what's going on.

BeeWarloc:
Now it should be ready.

src/runtime/Uno.AppLoader-MonoMac/Uno.AppLoader-MonoMac.csproj
```diff
@@ -108,9 +111,9 @@
        Other similar extension points exist, see Microsoft.Common.targets.
   <Target Name="BeforeBuild">
   </Target> -->
-  <Target Name="AfterBuild">
+  <!-- <Target Name="AfterBuild">
     <Exec Command="clang -m64 monostub.m -o bin/$(Configuration)/monostub -framework AppKit -mmacosx-version-min=10.6" />
-  </Target>
+  </Target> -->
```
kusma: Maybe we should just delete this instead of commenting it out?

src/runtime/Uno.AppLoader-MonoMac/monostub.m
```diff
@@ -1,4 +1,6 @@
-//clang -m64 monostub.m -o monostub -framework AppKit -mmacosx-version-min=10.6
+// We include a precompiled monostub to make compilation of solution possible on non-mac platforms,
+// so if this file is modified, it needs to be manually recompiled:
+//     clang -m64 monostub.m -o monostub -framework AppKit -mmacosx-version-min=10.6
```
kusma: Maybe this should be added as `src/runtime/Uno.AppLoader-MonoMac/README.md` instead, to increase visibility? Just a thought...

BeeWarloc: I can add a `README.md` file too. Howver, it's not given that someone trying to change `monostub.m` will read the README, so keeping the comment to avoid any surprises.

src/runtime/Uno.AppLoader-MonoMac/Uno.AppLoader-MonoMac.csproj
BeeWarloc: 👍

kusma:
Awesome :)

PR #28, opened 2018-05-23
Update Fuse to use sdk: 26 by default

ckarmy:
Google are going to **stop accepting apps** to the store which target at least Android Oreo.

**August 2018:** New apps required to target API level 26 (Android 8.0) or higher.
**November 2018:** Updates to existing apps required to target API level 26 or higher.
2019 onwards: Each year the targetSdkVersion requirement will advance. Within one year following each Android dessert release, new apps and app updates will need to target the corresponding API level or higher.

More info [here](https://github.com/fuse-open/fuselibs/issues/1187)!

### DIFF
--- a/Library/Core/UnoCore/Targets/Android/Android.uxl
+++ b/Library/Core/UnoCore/Targets/Android/Android.uxl
@@ -18,7 +18,7 @@
     <Set JDK.Directory="@(Config.Java.JDK.Directory:Path || JAVA_HOME:Env)" />
     <Set NDK.Directory="@(Config.Android.NDK.Directory:Path || ANDROID_NDK:Env)" />
     <Set NDK.PlatformVersion="@(Project.Android.NDK.PlatformVersion || Config.Android.NDK.PlatformVersion || '9')" />
-    <Set SDK.BuildToolsVersion="@(Project.Android.SDK.BuildToolsVersion || Config.Android.SDK.BuildToolsVersion || '26.0.1')" />
+    <Set SDK.BuildToolsVersion="@(Project.Android.SDK.BuildToolsVersion || Config.Android.SDK.BuildToolsVersion || '26.0.2')" />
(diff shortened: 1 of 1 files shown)


Library/Core/UnoCore/Targets/Android/Android.uxl
```diff
@@ -18,11 +18,11 @@
     <Set JDK.Directory="@(Config.Java.JDK.Directory:Path || JAVA_HOME:Env)" />
     <Set NDK.Directory="@(Config.Android.NDK.Directory:Path || ANDROID_NDK:Env)" />
     <Set NDK.PlatformVersion="@(Project.Android.NDK.PlatformVersion || Config.Android.NDK.PlatformVersion || '9')" />
-    <Set SDK.BuildToolsVersion="@(Project.Android.SDK.BuildToolsVersion || Config.Android.SDK.BuildToolsVersion || '23.0.1')" />
-    <Set SDK.CompileVersion="@(Project.Android.SDK.CompileVersion || Config.Android.SDK.CompileVersion || '23')" />
+    <Set SDK.BuildToolsVersion="@(Project.Android.SDK.BuildToolsVersion || Config.Android.SDK.BuildToolsVersion || '26.0.1')" />
```
kusma: Hmm, are we sure we want 26.0.*1*?

kusma:
@ckarmy: Would you mind signing the CLA? We might get rid of it at some point, but not right now...

Library/Core/UnoCore/Targets/Android/Android.uxl
kusma: Seems like the latest version of 26.0 is [26.0.2](https://developer.android.com/studio/releases/build-tools), perhaps we should use that version instead?

kusma:
@ckarmy: Ah, seems there's some issue with the git author information for the first commit, so it's not detected as the same github user... Perhaps `git commit --amend --reset-author` during a rebase will solve it?

PR #31, opened 2018-05-25
Uno.UX.Markup.*: fix casing of System.Xml

kusma:
This is not named System.XML, as we used here. This worked without
this patch, because we only build on Windows and case-insensitive
HFS+ currently.

This patch opens the door for case-sensitive HFS+, as well as Linux.


kusma:
Again, just for reference: This patch on top allows me to build somewhat cleanly on Linux:
```patch
diff --git a/scripts/build.sh b/scripts/build.sh
index dbc6b5e9..765e0f11 100755
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -94,7 +94,7 @@ if [ "$BUILD_PLATFORM" = 1 ]; then
 
         ;;
 
-    darwin*)
+    *)
         # Build C# solution
         csharp-build uno-macos.sln
 
```

... but I'm not going to submit that yet, as we should rather unify the solution files first, and this need will go away.

PR #32, opened 2018-05-25
Fix App Store issue while uploading the app to App Store because of App Icons are invalid format 

devadiab:
Fix App Store issue while uploading the app to App Store because of App Icon has invalid format and they have to be stripped from alpha channel.

The previous image was generated using default settings of .NET graphics. I have fixed it and adding additional app icon sizes by simply setting `CompositingMode` attribute for .NET graphics to `CompositingMode.SourceOver`

for further information see this example of XCode error message after archiving the App

Xcode error screen

Also it's listed here in Apple's guidelines for [Human Interface Guidelines ](https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/app-icon/):

> Keep the background simple and avoid transparency. Make sure your icon is opaque, and don’t clutter the background. Give it a simple background so it doesn’t overpower other app icons nearby. You don’t need to fill the entire icon with content.

### DIFF
--- a/src/compiler/Uno.Compiler.Core/Syntax/UxlProcessor.cs
+++ b/src/compiler/Uno.Compiler.Core/Syntax/UxlProcessor.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Drawing;
+using System.Drawing.Imaging;
 using System.Drawing.Drawing2D;
 using System.IO;
 using System.Linq;
@@ -619,10 +620,16 @@ namespace Uno.Compiler.Core.Syntax
 
                 using (var originalImg = Image.FromFile(src))
                 {
+                    if (_env.IsDefined("iOS") && Image.IsAlphaPixelFormat(originalImg.PixelFormat) )
+                    {
+                        var source = new Source(src);
+                        Log.Warning(source, ErrorCode.E0000, "iOS App Store doesn't accept Images with transparency.");
+                    }
+                  
                     int width = f.TargetWidth ?? f.TargetHeight ?? originalImg.Width;
                     int height = f.TargetHeight ?? f.TargetWidth ?? originalImg.Height;
 
-                    using (var resizedImg = new Bitmap(width, height))
+                    using (var resizedImg = new Bitmap(width, height, originalImg.PixelFormat))
                     {
                         using (var graphics = Graphics.FromImage(resizedImg))
                         {


src/compiler/Uno.Compiler.Core/Syntax/UxlProcessor.cs
```diff
@@ -628,6 +629,11 @@ void WriteImageFile(ImageFile f)
                         {
                             graphics.InterpolationMode = InterpolationMode.HighQualityBicubic;
                             graphics.PixelOffsetMode = PixelOffsetMode.HighQuality;
+                            if (_env.GetString("Target").ToUpper() == "IOS")
+                            {
+                                graphics.Clear(Color.White);
+                                graphics.CompositingMode = CompositingMode.SourceOver;
+                            }
```
kusma: So, this is better than the previous attempt, but I'm still not really convinced. As I [wrote here](https://github.com/fuse-open/uno/pull/30#issuecomment-391974467):

>>> This makes sure that if the input-image has an alpha channel, the output image will also. This leaves the "should my icon have an alpha-channel"-question up to the user, rather than a hard decision by us...

>> Apple's policy obligate us to do that, because it's not accepting the app to be uploaded to iTunesConnect even for TestFlight if the icon image has alpha channel !

> I don't think that's quite right. Our default icon does not have an alpha-channel AFAIK, and if the user has specified a custom icon with an alpha-channel, this is a user-error IMO, which can easily be fixed in the user-project. So I don't think we need to handle the case where the image has an alpha-channel other than forwarding that alpha channel.
>
> Removing the alpha channel, by blending the image on top of a background color makes assumptions about what the desired background color is, which is rather arbitrary. I'd rather not make arbitrary decisions like that.

What exactly are you trying to fix here? It sounds to me like the specified icon does indeed have translucency in it. Would it be better to fix the input-assets than to batch-process them with an arbitrary color here?

Library/Core/UnoCore/Targets/iOS/@(Icons)/Contents.json
```diff
@@ -12,6 +12,12 @@
       "size" : "20x20",
       "scale" : "3x"
     },
+    {
+      "filename" : "iPhone_29_1x.png",
+      "idiom" : "iphone",
+      "size" : "29x29",
+      "scale" : "1x"
+    },
```
kusma: I don't think any supported device has this icon size...

Library/Core/UnoCore/Targets/iOS/@(Icons)/Contents.json
```diff
@@ -35,6 +41,18 @@
...
+      "filename" : "iPhone_57_1x.png",
+      "idiom" : "iphone",
+      "size" : "57x57",
+      "scale" : "1x"
+    },
+    {
+      "filename" : "iPhone_57_2x.png",
+      "idiom" : "iphone",
+      "size" : "57x57",
+      "scale" : "2x"
```
kusma: I don't think any supported device has these icon sizes...

Library/Core/UnoCore/Targets/iOS/@(Icons)/Contents.json
```diff
@@ -84,6 +102,18 @@
...
+      "idiom" : "ipad",
+      "size" : "50x50",
+      "scale" : "1x"
+    },
+    {
+      "filename" : "iPad_50_2x.png",
+      "idiom" : "ipad",
+      "size" : "50x50",
+      "scale" : "2x"
+    },
```
kusma: I don't think any supported device has these icon sizes...

kusma:
[Here's some code](https://gist.github.com/kusma/6e554aed947b0905761491a932f61243) trying to show my point.

Now, if I run this project, I get the following output:
```
input: Format24bppRgb, output: Format32bppArgb
input: Format32bppArgb, output: Format32bppArgb
```
OK, so this is obviously not what we want, because we all of a sudden invented an alpha-channel in the first case! OK, so it might have all opaque pixels, but still. Not good.

However, if I pass `originalImg.PixelFormat` to the constructor for `Bitmap`, I get this instead:
```
input: Format24bppRgb, output: Format24bppRgb
input: Format32bppArgb, output: Format32bppArgb
```

OK, that's much better. Now the user can affect the output, and fix their projects.

I've also inspected the written files using TweakPNG, and it seems they have the format I expect them to...

src/compiler/Uno.Compiler.Core/Syntax/UxlProcessor.cs
kusma: [Here's some code](https://gist.github.com/kusma/6e554aed947b0905761491a932f61243) trying to show my point.

Now, if I run this project, I get the following output:
```
input: Format24bppRgb, output: Format32bppArgb
input: Format32bppArgb, output: Format32bppArgb
```
OK, so this is obviously not what we want, because we all of a sudden invented an alpha-channel in the first case! OK, so it might have all opaque pixels, but still. Not good.

However, if I pass `originalImg.PixelFormat` to the constructor for `Bitmap`, I get this instead:
```
input: Format24bppRgb, output: Format24bppRgb
input: Format32bppArgb, output: Format32bppArgb
```

OK, that's much better. Now the user can affect the output, and fix their projects.

I've also inspected the written files using TweakPNG, and it seems they have the format I expect them to...

devadiab: Okay let me try this, and get back.

Its good idea

devadiab: We have to agreed that there is a here 
https://github.com/fuse-open/uno/blob/master/src/compiler/Uno.Compiler.Core/Syntax/UxlProcessor.cs#L625
right?
And it still needs a contribution right ?
So while app says :
> *Keep the background simple and avoid transparency*. Make sure your icon is opaque, and don’t clutter the background. Give it a simple background so it doesn’t overpower other app icons nearby. You don’t need to fill the entire icon with content.

So the generated icon must avoid the transparency. so what is the best practice here ?

devadiab: Initially, `originalImg.PixelFormat` make things better when the input image is correct and fine and the code now preserves the situation of the image.
But should we raise an error event if the image have transparency ?

devadiab: I have full feedback:
1- If we use png image without transparency it will not work unless we add `originalImg.PixelFormat` or use `CompositingMode.SourceOver`.
2- if we use ping image with transparency it will not work at all.
3- We have to warn the use that the image contains transparency and fail the build.

Is that good? so i push a new commit ?

src/compiler/Uno.Compiler.Core/Syntax/UxlProcessor.cs
```diff
@@ -619,10 +620,19 @@ void WriteImageFile(ImageFile f)
 
                 using (var originalImg = Image.FromFile(src))
                 {
+
+                    if (_env.GetString("Target").ToUpper() == "IOS")
+                    {
+                        if (IfImageContainsTransparent((Bitmap)originalImg))
```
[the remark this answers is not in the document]
devadiab: [1] Done
[2] After investigation, Bitmap holds all types.
[3] Yes you are right, no need for my previous check.

src/compiler/Uno.Compiler.Core/Syntax/UxlProcessor.cs
```diff
@@ -619,10 +620,16 @@ void WriteImageFile(ImageFile f)
 
                 using (var originalImg = Image.FromFile(src))
                 {
+                    if (_env.IsDefined("iOS") && Image.IsAlphaPixelFormat(originalImg.PixelFormat) )
+                    {
+                        var source = new Source(src);
+                        Log.Warning(source, ErrorCode.E0000, "iOS App Store doesn't accept Images with transparency.");
```
kusma: Now you're no longer calling `IfImageContainsTransparent()` here... I think that's completely fine, but then we should also drop adding the `IfImageContainsTransparent()`-method...

devadiab: Removed and everything was squashed

src/engine/Uno.ProjectFormat/PropertyDefinitions.cs
```diff
@@ -115,4 +115,4 @@ public void Add(string property, PropertyType type, object defaultValue = null)
                 defaultValue?.ToString()));
         }
     }
-}
+}
```
kusma: You can probably revert this also.

Library/Core/UnoCore/Targets/iOS/iOS.uxl
```diff
@@ -135,4 +135,4 @@
     <ProcessFile Name="@(Schemes)/@(Project.Name).xcscheme" Overwrite=false />
     <ProcessFile Name="@(Schemes)/xcschememanagement.plist" Overwrite=false />
 
-</Extensions>
+</Extensions>
```
kusma: You can probably revert this also.

devadiab:
@kusma , i think we can go now ;)

PR #35, opened 2018-05-25
Three additional `Unoproj` Properties to help the Foreign code developers for iOS

devadiab:
I feel sick while everytime building the iOS app because i have to add some changed manually to the project. Those properties i used to integration with `PhotoEditorSDK` library and it requires the following actions each time:

1- Allow Embed Swift Standard Libraries. 
2- Allow non-modular includes in Frameworks Modules.
3- Adding two flags to Apple's LLVM Compiler to fix the following compiler error `Use of '@import' when C++ modules are disabled, consider using -fmodules and -fcxx-modules`, So i added `-fmodules` and `-fcxx-modules` flags to it.

Those suggested attributes are nessesary for anyone need to connect with Swift SDK and Swift-NDK modules.

And the attributes are :
1- AllowSwiftModules
2- AllowNonModularModules
3- CPPImport

And can be used for example :
```
 "iOS": {
    "SpecialCapabilities": {
      "AllowSwiftModules": true,
      "AllowNonModularModules": true,
      "CPPImport": true
    },
```

That's all ;)
Thanks

### DIFF
--- a/Library/Core/UnoCore/Targets/iOS/@(Project.Name).xcodeproj/project.pbxproj
+++ b/Library/Core/UnoCore/Targets/iOS/@(Project.Name).xcodeproj/project.pbxproj
@@ -458,10 +458,31 @@
 #if @(LIBRARY:Defined)
                 MACH_O_TYPE = mh_dylib;
 #endif
+
+#if @(Project.iOS.SpecialCapabilities.CPPImport:IsSet)
+    #if @(Project.iOS.SpecialCapabilities.CPPImport:Test(1,0))
+                OTHER_CPLUSPLUSFLAGS = (
+					"$(OTHER_CFLAGS)",
+					"-fmodules",
+					"-fcxx-modules",
+				);
+    #endif
+#endif
+#if @(Project.iOS.SpecialCapabilities.AllowSwiftModules:IsSet)
+    #if @(Project.iOS.SpecialCapabilities.AllowSwiftModules:Test(1,0))
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
+    #endif
+#endif
+#if @(Project.iOS.SpecialCapabilities.AllowNonModularModules:IsSet)
+    #if @(Project.iOS.SpecialCapabilities.AllowNonModularModules:Test(1,0))
+                CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES;
+    #endif
+#endif
                 OTHER_LDFLAGS = (
                     "$(inherited)",
                     "@(Pbxproj.LinkLibraries)"
                 );
+
                 PRODUCT_NAME = "$(TARGET_NAME)";
                 SWIFT_VERSION = @(Project.iOS.SwiftVersion || "3.0");
             };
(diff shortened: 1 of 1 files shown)


Library/Core/UnoCore/Targets/iOS/@(Project.Name).xcodeproj/project.pbxproj
```diff
@@ -458,10 +458,31 @@
 #if @(LIBRARY:Defined)
                 MACH_O_TYPE = mh_dylib;
 #endif
+
+#if @(Project.iOS.SpecialCapabilities.CPPImport:IsSet)
```
kusma: I'm not so fond of the `SpecialCapabilities`-namespace here. It seems a bit odd IMO; the rest of the settings under `Project.iOS` are just as "special", no?

Library/Core/UnoCore/Targets/iOS/@(Project.Name).xcodeproj/project.pbxproj
```diff
@@ -458,10 +458,31 @@
...
+#if @(Project.iOS.SpecialCapabilities.AllowSwiftModules:IsSet)
+    #if @(Project.iOS.SpecialCapabilities.AllowSwiftModules:Test(1,0))
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
+    #endif
+#endif
+#if @(Project.iOS.SpecialCapabilities.AllowNonModularModules:IsSet)
+    #if @(Project.iOS.SpecialCapabilities.AllowNonModularModules:Test(1,0))
+                CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES;
+    #endif
+#endif
```
kusma: ...but another question is, do we really need to teach the uno compiler about every possible build-setting? Wouldn't it be easier to add something like 

@(Project.iOS.CustomBuildSettings:Join('\n'))

...so the user can specify these themselves without us having to approve every setting?

Library/Core/UnoCore/Targets/iOS/@(Project.Name).xcodeproj/project.pbxproj
devadiab: This namespace proposed by me, we can pick the better name ;)

Library/Core/UnoCore/Targets/iOS/@(Project.Name).xcodeproj/project.pbxproj
devadiab: You're right, but it's highly recommended to let this go as we need to ensure that xcode project file will be correct and fine.
By this way, we don't need to have extra validation for the rules because it predefined strong value either (true) or (false).
Please advice 👍

Library/Core/UnoCore/Targets/iOS/@(Project.Name).xcodeproj/project.pbxproj
kusma: I think just `Project.iOS.SpecialCapabilities.CPPImport` is good enough...

Library/Core/UnoCore/Targets/iOS/@(Project.Name).xcodeproj/project.pbxproj
kusma: My knee-jerk reaction is that I think I'd prefer throwing a wider net, by providing a weaker but generally more useful mechanism.

That being said, I totally see the point in having strong, less error-prone switches. I'm not really sure what the right balance is here...

devadiab: what we should do right now, @kusma  ?
>I continue creating PRs from our development environments to match all Fuse developers needs and we have advance and larger apps day by day.

kusma: @devadiab: I'm willing to merge either this (but perhaps with the ususual SpecialCapabilities-namespace removed), or a more generic (but "unsafe") solution.

kusma: I also suspect that we should add some documentation somewhere?

devadiab: About namespace, take your time and pick a good one okay?
And we can create another PRs for documentation or i can create another commit with documentation. seems good ?

kusma: @devadiab: What I've been trying to hint at already (perhaps not clearly enough), is that I think directly in the `Project.iOS.*`-namespace would be fine. That would mean, for example `Project.iOS.CPPImport`.

devadiab: Why not, its a good idea; let me change

kusma: @devadiab: Any chance of this happening? Seems you haven't pushed out such a change yet...

kusma:
OK, I'm going to close this for now.

@devadiab: If you decide to finish this up, feel free to reopen :)

PR #40, opened 2018-06-02
Fix android permissions crash while _currentRequest is null

devadiab:
I have reformated the braces and fixed the `NullPointerException` crash by copying the condition used on `nextRequest()` method.

I figured it out by using `Firabase Crashlytics` and it says that there is a crash with the following information: 

```
Caused by java.lang.NullPointerException
Attempt to read from field 'int com.fuse.PermissionsRequest.requestID' on a null object reference
```
```
com.fuse.Permissions.onRequestPermissionsResult (Permissions.java:80)
```
Best Regards.

### DIFF
--- a/Library/Core/Uno.Permissions/Permissions.java
+++ b/Library/Core/Uno.Permissions/Permissions.java
@@ -76,7 +76,11 @@ public final class Permissions {
             _currentRequest.requestID);
     }
 
-    public static void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+    public static void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) 
+    {
+        if( _currentRequest == null) 
+            return;
+
         if (_currentRequest.requestID == requestCode && _currentRequest.promise != null && grantResults.length > 0) 
         {
             boolean ok = true;


Library/Core/Uno.Permissions/Permissions.java
```diff
@@ -76,7 +76,11 @@ private static void nextRequest()
             _currentRequest.requestID);
     }
 
-    public static void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+    public static void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) 
+    {
+        if( _currentRequest != null || _requests.size() == 0)
```
kusma: Hmm, I'm not sure if we need that `_requests.size() == 0`-check... The code below doesn't seem to depend on this variable at all...

devadiab: Humm, I see .

It changed now, thanks for your review

Library/Core/Uno.Permissions/Permissions.java
```diff
@@ -76,7 +76,11 @@ private static void nextRequest()
             _currentRequest.requestID);
     }
 
-    public static void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+    public static void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) 
+    {
+        if( _currentRequest != null)
```
mortend: To me, it looks like you want the early-out when `_currentRequest == null`.

As it stands now, the method will still crash when `_currentRequest` is null, and just do nothing when not null...

devadiab: You are right, thanks for your advice ;)

devadiab: It fixed now @mortend

mortend: Nice :)

kusma: Whoops, sloppy review on my end! Good catch!

PR #42, opened 2018-06-02
Uno.CLI: catch exceptions to avoid hanging in LaunchApk

mortend:
Sometimes we get exceptions inside the timers, and we don't want our program to
get stuck..

Specifically, I just got an exception from AdbDevice.GetProcessId().

### DIFF
--- a/src/main/Uno.CLI/Android/LaunchApk.cs
+++ b/src/main/Uno.CLI/Android/LaunchApk.cs
@@ -94,8 +94,17 @@ namespace Uno.CLI.Android
                 timer.Elapsed += (s, e) => {
                     timer.Stop();
                     Shell.KillAll();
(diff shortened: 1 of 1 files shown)


mortend:
It does fix something :)

If we get an exception inside the timer we never reach the `Environment.Exit()` calls, and get stuck in a loop.

I had an Android app crashing at startup, so launch-apk failed to get the process ID, causing an exception, and then got stuck at a screen like this forever:
```
Installing APK on 1 device(s)
Launching activity 'app'
```

After the fix the program now prints `Process 'com.apps.app' failed to start` and terminates.

mortend:
I also added a second `try..catch` in the last timer, which wasn't really necessary in my case, but better safe than sorry I guess..

PR #48, opened 2018-06-04
UnoCore: add ValueType class

mortend:
This fixes two problems when compiling on .NET and Mono:

1) BytecodeCompiler compiles `base` wrong, causing invalid program exceptions
   on both platforms.

2) Calling `object.Equals()` on value types gives unexpected result on Mono,
   causing assertion failures.

After splitting up virtual methods on `object`, to `Enum` and `ValueType`, we get rid of some if-tests on C++, which should give slighly faster calls to `GetHashCode`, `Equals` and `ToString` as a bonus.


tests/src/UnoTest/General/ValueTypes.uno
```diff
@@ -0,0 +1,38 @@
...
+        public override bool Equals(object o)
+        {
+            return base.Equals(o);
+        }
+    }
+
+    class ValueTypes
+    {
+        [Test]
+        public static void Test()
```
kusma: Could you mark this as `[Ignore]` until the test will pass? That way, we don't get broken bisections past this point...

mortend: Sure, hold on!

mortend: Done.

PR #49, opened 2018-06-04
Uno.Compiler.Frontend: early-out in TryParseCondition()

mortend:
This avoids false calls to `Log.Error()`, which is a convenient place to set
breakpoints. So, let's make life a little bit easier when debugging the
compiler..


mortend:
When compiling for Android, I had some macros expanding to empty strings, i.e.:
```
#if @(Foo)
```
And then I got several breaks in `Log.Error()` because `Foo` wasn't set. I don't remember all the details..

PR #51, opened 2018-06-04
Uno.Compiler: report error on unsupported 'extern' code

mortend:
If we try to compile 'extern' code on .NET we'll get exceptions in `CilBackend`, i.e.:
```
void foo()
@{
    // code here
@}
```
These errors are hard to fix because the compiler doesn't give us any source information.

After this fix, we get a more helpful error message that tells us where to fix the problem.

### DIFF
--- a/Library/Core/UnoCore/Source/OpenGL/GL.uno
+++ b/Library/Core/UnoCore/Source/OpenGL/GL.uno
@@ -764,9 +764,9 @@ namespace OpenGL
                 glBindTexture($@);
             @}
             else if defined(DOTNET)
-            @{
-                _gl.BindTexture($@);
-            @}
+            {
+                _gl.BindTexture(target, texture);
+            }
             else
                 build_error;
         }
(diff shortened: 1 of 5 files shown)


Library/Core/UnoCore/Source/OpenGL/GL.uno
```diff
@@ -764,9 +764,9 @@ namespace OpenGL
                 glBindTexture($@);
             @}
             else if defined(DOTNET)
-            @{
-                _gl.BindTexture($@);
-            @}
+            {
+                _gl.BindTexture(target, texture);
+            }
```
kusma: I think this change should come first, to avoid bisect issues...

mortend:
Just some more clarifications:

1. The patch doesn't introduce any new errors, it changes a cryptic exception into a human-readable error message.

2. The UnoCore patch doesn't change anything at this point, because the code has gone through the C#-backend and gets linked in via `Uno.Runtime.Core.dll` by the CIL-backend, so the actual code ends up being ignored.

I discovered these issues while removing the C#-backend and `Uno.Runtime.Core.dll` on a fork.

PR #55, opened 2018-06-13
Fixup UnoCore generation

kusma:
Re-generating UnoCore is not currently possible, due to a few mismatches between the current sources and the generator. This isn't ideal, as it means people will have to carefully stage only the changes they want.

This series have two small fixes "tacked on", but they are all kinda related:
- The FirefoxOS-define is never defined any more, so checking for it is kinda pointless.
- The C# writer now emits `new`-keywords for definitions that previously missed them. This eliminates a compile-time warning.


kusma:
Just thought I'd mention: I've verified on Mono (on Linux, I don't have a mac any more) and Windows that this generates the same code.

PR #59, opened 2018-06-19
ILVerifier: check foreign attribute in VerifyExtern()

mortend:
This fixes build errors when building fuselibs.


mortend:
The regression was introduced in #51

PR #60, opened 2018-06-19
Test transform improvements

mortend:
Some improvements to TestSetupTransform in the compiler.

I just ran the fuselibs test suite to make sure things still work.

### DIFF
--- a/src/compiler/Uno.Compiler.Core/IL/Testing/TestSetupTransform.cs
+++ b/src/compiler/Uno.Compiler.Core/IL/Testing/TestSetupTransform.cs
@@ -95,10 +95,15 @@ namespace Uno.Compiler.Core.IL.Testing
                                           InvokeName(method),
                                           DataType.Void,
                                           ParameterList.Empty);
+            var testVar = new Variable(_source, invokeMethod, "obj", method.DeclaringType,
+                                       VariableType.Default, InstantiateTestFixture(method));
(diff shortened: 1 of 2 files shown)


src/compiler/Uno.Compiler.Core/IL/Testing/TestSetupTransform.cs
```diff
@@ -88,11 +88,25 @@ public void CreateTestSetUpClass()
...
+            if (method.IsStatic || ctor.IsGenerated && !method.DeclaringType.IsImplementingInterface(Essentials.IDisposable))
+                return method;
+
             var invokeMethod = new Method(_source,
                                           _mainClass,
                                           null,
-                                          Modifiers.Generated | Modifiers.Public | Modifiers.Static,
-                                          InvokeName(method),
+                                          Modifiers.Generated | Modifiers.Private | Modifiers.Static,
+                                          method.FullName.Replace('.', '_').ToIdentifier(),
```
kusma: Doesn`t this increase the risk name collisions? I think that's why we added the invoke-prefix. I'm not saying that's robust at all, but it seems slightly better than dropping it...

kusma: I don't think I follow, but since the old code wasn't exactly iron-clad either... why not? :)

PR #61, opened 2018-06-19
Thread cleanup

mortend:
This cleans up the `Thread` implementation.

We also move the `Thread` class to `UnoCore`, as I think it's essential. Because of this, we no longer depend on `Uno.Threading` in the `Uno.Testing` package.

Also, the new C++ TLS API makes us depend less on `uno-base`.


Library/Core/UnoCore/Backends/CPlusPlus/Uno/Support.cpp
```diff
@@ -91,6 +97,47 @@ void uFatal(const char* src, const char* msg)
     abort();
 }
 
+uThreadLocal* uCreateThreadLocal(void (*destructor)(void*))
+{
+#ifdef WIN32
+    // TODO: Handle destructor...
```
mortend: The old implementation in `uno-base` didn't handle this either, meaning we still get a small mem-leak per thread on Win32/C++ builds. I don't think this is a major issue at this point, since it's not a regression, and this backend isn't much used.

Library/Core/UnoCore/Source/Uno/Threading/Thread.uno
```diff
@@ -0,0 +1,183 @@
...
+            extern "uSetThreadLocal(@{_currentThread}, $0)";
+
+            try
+            {
+                thread._threadStart();
+            }
+            catch (Exception e)
+            {
+                // TODO: Use some kind of exception callback..
+                debug_log "Unhandled exception in thread: " + e;
```
mortend: Previously, this would cause an unhandled C++ exception on the running thread (and leak the Thread-object).

Library/Core/UnoCore/Source/Uno/Threading/Thread.uno
```diff
@@ -0,0 +1,183 @@
...
+            // This is a bit tricky. We need to retain *before* the thread starts,
+            // otherwise the thread might be deleted before it has started.
+            // To avoid leaks, ThreadMain() will release the thread.
+            extern "uRetain($$)";
+
+            if defined(UNIX)
+            {
+                if (extern<int> "pthread_create(&@{$$._threadHandle}, NULL, _ThreadFunc, (void*)$$)" != 0)
+                {
+                    extern "uRelease($$)";
```
mortend: Previously this would leak the Thread-object, but now we've added code to release it.

Library/Core/UnoCore/Source/Uno/Threading/Thread.uno
```diff
@@ -0,0 +1,183 @@
...
+                    throw new InvalidOperationException("pthread_create() failed!");
+                }
+            }
+            else if defined(WIN32)
+            {
+                extern "@{$$._threadHandle} = ::CreateThread(NULL, 0, _ThreadFunc, (LPVOID)$$, 0, NULL)";
+
+                if (extern<bool>(_threadHandle) "!$0" )
+                {
+                    extern "uRelease($$)";
```
mortend: Previously this would leak the Thread-object, but now we've added code to release it.

mortend:
I don't think that's available on all our C++ backends, at least not last time I tried... Ran into some issues!

mortend:
See for example the `#if IOS` stuff in https://github.com/fuse-open/uno/blob/master/Library/Core/Uno.Threading/Implementation/Posix/posix_thread.cpp

kusma:
@mortend: That's about non standard-mandated pthreads functionality. I think we have C++11 on all targets now.

mortend:
Also, we're still on VS2013 on `WIN32` where I'm pretty sure we'll get issues because VS2013 only has partial C++11 support. If I remember correctly that's where I gave up, and decided it was easier to keep using the platform specific APIs we were already using.

This is starting to become a long time ago though, so I don't remember the details, and I don't actually have VS2013 on any of my current computers anymore...

Anyway, on my fork, I've upgraded to VS2017 compilers in this patch: https://github.com/mortend/uno/commit/bd318dbf348f67b92a1c9c6cadbab8c47a7f700a. Do you think we should make the effort to upstream the patch, and recompile our dependencies for VS2017? If so, I'll need help with the dependencies.. 👍

kusma:
Hmm. There's been a long time since I even had vs2013 installed, things have just been working for me, I believe.

I think modernizing the msvc usage is a good idea; we already tried to standardize on 2017 before we open sourced everything.

I don't know if I'll have time to help out with updating dependencies, but maybe this PR is a good first step regardless?

mortend:
Hmm, yeah - I think this PR is a good first step and unrelated to the VS2017 discussion. We can refactor further to C++11 threads in an upcoming PR, and I think this PR should make that job a little bit easier.

Regarding VS2017 support: I can provide my VS2017 patch, and build `uno-core` binaries for VS2017, but I don't have a place to host them. Do we still have access to `files.fusetools.com`? I haven't checked what needs to be done on the fuselibs side, though... Could be nothing if we're lucky, but I don't know!

kusma:
@mortend: No, I don't have access to `files.fusetools.com` any more... :(

Ideally, we should switch stuff to being hosted via nuget on our myget-stream, but that might be a lot of work.

mortend:
@kusma: I see.

Ah, yes, we can upload a hidden NuGet package - this doesn't need to be a lot of work, I think. :)

mortend:
See #62.

Library/Core/Uno.Threading/Implementation/Posix/posix_thread.cpp
```diff
@@ -34,23 +26,13 @@ static void* ThreadStartup(void* arg)
     uDelegate* threadStart = @{Uno.Threading.Thread:Of(thread)._threadStart:Get()};
     assert(threadStart != NULL);
 
-#if IOS
     EnsureInitialized();
     pthread_setspecific(currentThread, (void*)thread);
```
kusma: I believe we had bad crashes with this on old iOS devices. Something about pthread_setspecific not being supported on all iOS versions. It's only 1.5 years since that fix was introduced, so I'm a bit skeptical to changing it just like that...

This is 72ad247c07c918e23119ba8cf58638912859a3db in the old commit history (if you have access to that). I couldn't find a PR nor issue, but that might just be me being bad at searching.

mortend: The code that was previously special-cased for iOS is now what we use on all Unix-platforms, a general implementation using pthread specifics instead of the `__thread` storage specifier for TLS.

I believe it must be the `__thread`-based code that caused those crashes, since we were using that on non-iOS Unix-platforms.

So this PR won't reintroduce bad crashes on iOS since the code is still the same.

mortend: We're now using pthread specifics in the new cross-platform C++ wrapper: https://github.com/mortend/uno/blob/274644401421d7e35d07c0d127f2c0822179d2fc/Library/Core/UnoCore/Backends/CPlusPlus/Uno/Support.cpp#L128

The cross-platform wrapper is used by Uno in the static constructor here (previously `EnsureInitialized()`): https://github.com/mortend/uno/blob/274644401421d7e35d07c0d127f2c0822179d2fc/Library/Core/UnoCore/Source/Uno/Threading/Thread.uno#L44

We need the cross-platform wrapper since we also use it in `Memory.cpp` to handle auto-release pools etc.

PR #62, opened 2018-06-21
Upgrade to VS2017 C++ compilers

mortend:
We now use VS2017 instead of VS2013 to build C++ code on WIN32 platforms.


mortend:
I had to rebuild Harfbuzz (with freetype2 and ICU-support) to get things going again on fuselibs.
PR and NuGet-package coming soon!

However, we'll need a new Uno-release containing VS2017-support for working CI on fuselibs.
No rush, but I suppose we're now almost ready for `release-1.10`...

PR #68, opened 2018-06-29
Android: upgrade Gradle versions

mortend:
This upgrades to Gradle 4.4, and Gradle-plugin 3.1.3.

This gives access to a new feature called "Instant Run", and we avoid being asked if we want to upgrade every time we open a generated project in Android Studio.

### DIFF
--- a/Library/Core/UnoCore/Targets/Android/Android.uxl
+++ b/Library/Core/UnoCore/Targets/Android/Android.uxl
@@ -18,7 +18,7 @@
     <Set JDK.Directory="@(Config.Java.JDK.Directory:Path || JAVA_HOME:Env)" />
     <Set NDK.Directory="@(Config.Android.NDK.Directory:Path || ANDROID_NDK:Env)" />
     <Set NDK.PlatformVersion="@(Project.Android.NDK.PlatformVersion || Config.Android.NDK.PlatformVersion || '9')" />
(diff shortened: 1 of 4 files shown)


Library/Core/UnoCore/Targets/Android/Uno/Graphics/GLHelper.cpp
```diff
@@ -240,7 +240,7 @@ void GLHelper::_setEGLConfig(bool forPBuffer)
         eglGetConfigAttrib(_eglDisplay, configs[i], EGL_SAMPLES, &samples);
 
         if (samples >= cs && depth >= cd && buffer >= cb &&
-            samples <= samples && r <= colorBits.R && g <= colorBits.G && b <= colorBits.B && a <= colorBits.A)
+            r <= colorBits.R && g <= colorBits.G && b <= colorBits.B && a <= colorBits.A)
```
kusma: Wow, this is actually a bug. There's two variables with the same name here, I bet the intention was to compare them.

That being said, I don't think we want to default to 4xMSAA any more, so maybe we should just get rid of that variable as well?

Anyway, this all seems a bit out-of-place here. I guess you noticed this due to some new warning when upgrading? Either way, I think this would deserve its own PR, unless it cause problems for this one without.

mortend: I noticed this because of warnings introduced by the new tools, yes.

Another PR coming up! :)

PR #69, opened 2018-06-29
Android: remove self-comparison in GLHelper.cpp

mortend:
This expression will always be true, so remove it to avoid compile-time warning on new build-tools.

### DIFF
--- a/Library/Core/UnoCore/Targets/Android/Uno/Graphics/GLHelper.cpp
+++ b/Library/Core/UnoCore/Targets/Android/Uno/Graphics/GLHelper.cpp
@@ -240,7 +240,7 @@ void GLHelper::_setEGLConfig(bool forPBuffer)
         eglGetConfigAttrib(_eglDisplay, configs[i], EGL_SAMPLES, &samples);
 
         if (samples >= cs && depth >= cd && buffer >= cb &&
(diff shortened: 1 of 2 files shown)


Library/Core/UnoCore/Targets/Android/Uno/Graphics/GLHelper.cpp
```diff
@@ -240,7 +240,7 @@ void GLHelper::_setEGLConfig(bool forPBuffer)
         eglGetConfigAttrib(_eglDisplay, configs[i], EGL_SAMPLES, &samples);
 
         if (samples >= cs && depth >= cd && buffer >= cb &&
-            samples <= samples && r <= colorBits.R && g <= colorBits.G && b <= colorBits.B && a <= colorBits.A)
+            r <= colorBits.R && g <= colorBits.G && b <= colorBits.B && a <= colorBits.A)
```
kusma: maybe remove the definition of `samples` as well?

mortend: The definition of `samples` is now gone. :)

mortend:
I threw in another warning-fix as well... :)

PR #73, opened 2018-07-01
Android gradle cleanup

mortend:
Just some small clean-ups in the gradle build script.

### DIFF
--- a/Library/Core/UnoCore/Targets/Android/app/build.gradle
+++ b/Library/Core/UnoCore/Targets/Android/app/build.gradle
@@ -9,7 +9,7 @@ dependencies {
     implementation 'com.android.support:support-v4:23.4.0'
     implementation 'com.android.support:appcompat-v7:26.0.2'
     implementation 'com.android.support:design:26.0.2'
-    @(Gradle.Dependency.Compile:Join('\n', 'implementation \'', '\''))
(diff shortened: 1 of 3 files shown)


Library/Core/UnoCore/Targets/Android/app/build.gradle
```diff
@@ -41,9 +34,6 @@ repositories {
         }
     }
 #endif
-#if @(Gradle.Model.Repository:IsRequired)
-    @(Gradle.Model.Repository:Join('\n'))
```
kusma: @cbaggers added this in 0cb3d6ed52ffc19624877959038a24b0e012aa5a from the old history. Perhaps he can explain why?

mortend:
I'm following up with some commits I had prepared for an upcoming PR, so we can properly deprecate `Gradle.Module.Repository`. If anyone is using this it will now print a deprecation message and forward the values to `Gradle.Repository`, which is the same thing.

Library/Core/UnoCore/Targets/Android/app/build.gradle
mortend: I've added a fix for potential `Gradle.Model.Repository` users in de594a3089b337d528420fcb2e7dfed64236c10b.

PR #75, opened 2018-07-02
Android: add Gradle.Dependency.NativeImplementation

mortend:
This feature will extract native libraries from downloaded AARs, so that we can
use them in our Uno/C++ code and make the NDK able to link them.

The extracted libraries are also added to search-paths in CMakeLists.txt, so we
can simply link them without worrying about actual paths.

I'm currently using this feature to build an app using the ARCore SDK, and it
would be cool if I could test the next version of Fuse Studio on my app...

### DIFF
--- a/Library/Core/UnoCore/Targets/Android/app/build.gradle
+++ b/Library/Core/UnoCore/Targets/Android/app/build.gradle
@@ -4,21 +4,48 @@ apply plugin: 'com.android.library'
 apply plugin: 'com.android.application'
 #endif
 
+// A configuration to mark AARs to extract native libraries from
+#if @(Gradle.Dependency.NativeImplementation:IsRequired)
+configurations { native_implementation }
+#endif
+
 dependencies {
     implementation fileTree(dir: 'src/main/libs', include: ['*.jar'])
     implementation 'com.android.support:support-v4:23.4.0'
     implementation 'com.android.support:appcompat-v7:26.0.2'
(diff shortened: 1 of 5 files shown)


Library/Core/UnoCore/Targets/Android/app/build.gradle
```diff
@@ -4,22 +4,47 @@ apply plugin: 'com.android.library'
 apply plugin: 'com.android.application'
 #endif
 
+// A configuration to mark AARs to extract native libraries from
+configurations { native_implementation }
```
kusma: Should this be guarded by `if @(Gradle.Dependency.NativeImplementation:IsRequired)`? I mean, the rest of this feature is...

mortend: I added the guards because I wanted to avoid any overhead of the `tasks.whenTaskAdded {}`-logic, and not add non-existing search-paths when the feature is not used.

I don't know what makes most sense TBH (guards or not), maybe the guards make things a bit more explicit?

kusma: I dunno, I have no idea if this line cause any trouble on its own... ;)

mortend: This shouldn't cause any trouble or significant overhead.

When thinking about it... Let's just add the last guard to make it explicit for the reader that these chunks of code are related. Hold on! :)

PR #80, opened 2018-07-20
Replace Buffer class

mortend:
This introduces a new `Bytes` class containing extension methods for manipulating `byte` arrays directly.

We can now use plain `byte[]`, which are more practical than creating `Buffer` objects when we want to work with byte data, and the interface is still pretty much identical.

I think this is a good deal, and patches for `fuselibs` are also ready: https://github.com/mortend/fuselibs/tree/replace-buffer

### DIFF
--- a/Library/Core/UnoCore/Backends/CPlusPlus/Uno/Support.h
+++ b/Library/Core/UnoCore/Backends/CPlusPlus/Uno/Support.h
@@ -70,14 +70,14 @@ unsigned int uCreateGLTexture(uImage::Texture* texData, bool generateMipmap = tr
 #define U_BUFFER_PTR(buffer) ((uint8_t*)(buffer)->_data->_ptr + (buffer)->_offset)
 #define U_BUFFER_SIZE(buffer) (buffer)->_sizeInBytes
 
-void uReverseBytes(uint8_t* ptr, size_t size);
+void DEPRECATED("This method will be removed in a future version") uReverseBytes(uint8_t* ptr, size_t size);
 
 template<class T>
-void uReverseBytes(T& ref) {
+void DEPRECATED("This method will be removed in a future version") uReverseBytes(T& ref) {
     uReverseBytes((uint8_t*)&ref, sizeof(T));
 }
 template<class T>
-T uLoadBytes(uint8_t* ptr, bool littleEndian) {
+T DEPRECATED("This method will be removed in a future version") uLoadBytes(uint8_t* ptr, bool littleEndian) {
     T result;
     memcpy(&result, ptr, sizeof(T));
 
@@ -87,7 +87,7 @@ T uLoadBytes(uint8_t* ptr, bool littleEndian) {
     return result;
 }
 template<class T>
-void uStoreBytes(uint8_t* ptr, T value, bool littleEndian) {
+void DEPRECATED("This method will be removed in a future version") uStoreBytes(uint8_t* ptr, T value, bool littleEndian) {
     if (!littleEndian)
         uReverseBytes(value);
 
--- a/Library/Core/UnoCore/Source/Uno/Buffer.uno
+++ b/Library/Core/UnoCore/Source/Uno/Buffer.uno
@@ -1,10 +1,10 @@
 using Uno.Compiler.ExportTargetInterop;
-using Uno.Runtime.Implementation;
 using Uno.Runtime.InteropServices;
 using Uno.IO;
 
 namespace Uno
 {
+    [Obsolete("Use extension methods on Uno.ByteArrayExtensions instead")]
     public sealed class Buffer
     {
         public static Buffer Load(BundleFile file)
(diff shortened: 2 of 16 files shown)


Library/Core/UnoCore/Source/Uno/Bytes.uno
```diff
@@ -0,0 +1,721 @@
+using Uno.Compiler.ExportTargetInterop;
+
+namespace Uno
+{
+    public static class Bytes
```
kusma: Hmm, I think `Uno.Bytes` is a bit too abstract of a name... could we use something like `Uno.ByteArrayExtensions` or something like that instead? `Bytes` sounds like something we *might* want to keep for something else in the future...

mortend: OK - better safe than sorry, and since these are extension methods the actual name of the class won't be used much. Updated to `ByteArrayExtensions` now. 👍

Library/Core/UnoCore/Backends/CPlusPlus/Uno/Support.h
```diff
@@ -42,8 +40,6 @@ uBase::Vector2i uInt2ToXliVector2i(const @{int2}& vec);
 
 @{float2} uFloat2FromXliVector2(const uBase::Vector2& vec);
 uBase::Vector2 uFloat2ToXliVector2(const @{float2}& vec);
-
-@{Uno.Buffer} uBufferFromXliDataAccessor(const uBase::DataAccessor* data);
```
kusma: I think we should keep these things around until `Uno.Buffer` has actually been removed, otherwise we break existing code...

Library/Core/UnoCore/Backends/CPlusPlus/Uno/Support.h
```diff
@@ -62,35 +58,3 @@ struct uGLTextureInfo
...
-
-    return result;
-}
-template<class T>
-void uStoreBytes(uint8_t* ptr, T value, bool littleEndian) {
-    if (!littleEndian)
-        uReverseBytes(value);
-
-    memcpy(ptr, &value, sizeof(T));
-}
```
kusma: Probably the same with these...

Library/Core/UnoCore/Backends/CPlusPlus/Uno/Support.h
mortend: The only user of this method was removed in https://github.com/fuse-open/uno/pull/80/commits/a5d165f2a939fb90c2650b11fdb2bb1f107bae30#diff-4aed3dbdda54ba506cf8d9e73348defcL91, prior to the commit that removes the method itself.

The method is also `uno-base` specific, not used by `fuselibs`, `Fuse.Entities` or `fuse-studio`, and I think it's very unlikely that any third-parties are using this.

Also, since `Support.h` must be included separately, it is not really part of the "core C++ API" I would say. I don't think it will be a problem to remove this.

Only `U_BUFFER_PTR` and `U_BUFFER_SIZE` were used by `Experimental.TextureLoader` in `fuselibs`, so those are kept for now. When https://github.com/mortend/fuselibs/tree/replace-buffer is merged, these macros won't be used anymore either...

Library/Core/UnoCore/Backends/CPlusPlus/Uno/Support.h
mortend: Our new `ByteArrayExtensions` class uses foreign code to inline the `memcpy` code, so we no longer have any users of these methods. The old `Buffer` class just forwards calls to `ByteArrayExtensions` now.

Library/Core/UnoCore/Backends/CPlusPlus/Uno/Support.h
kusma: I wouldn't be so sure. There's a lot of foreign code floating around...

kusma: OK, just let me fill in a bit why I worry about things like this. Experience from our release-management in the past has shown that some people tend to use code we have no idea about, and we don't find out about it until when it's way too late. The end result can be that the product looks unreliable, and that old, not actively maintained code tends to not work. Which is a problem for a product that has a lot of not-really-maintained 3rd-party out there... Just looking through the code that *we* see for compatibility issues has bitten us in the past.

You might be right in this particular case, it's just that the argument that this is unused because it's not used in fuselibs, Fuse.Entites and fuse-studio, well, that's been proven to not be enough before.

Now, I'm not proposing that we stop changing or removing APIs entirely, just that things like this should go through a normal deprecation cycle, and be documented as such in the changelog so people have a chance to catch up.

mortend: That's a thorough write-up, thank you! I understand your concerns, and in many cases I agree.

I'm sometimes a bit trigger happy when it comes to deleting old (and badly designed) code - the least risky route is of course to not delete the code, and also the least satisfying.

Anyway, here comes a rebase with the deprecations.

PR #91, opened 2018-08-20
update copyright info

mortend:
Fusetools AS no longer owns or maintains this project, so it's time to update this information.


mortend:
I see - I wanted to update this information before a new release goes out, but maybe it has to wait.

It looks like Facebook can claim copyright to the present, though: https://github.com/facebook/react-native/blob/master/LICENSE#L3

kusma:
Can, or tries to? When I took my OSI certification in software licensing, using "present" was never mentioned as an option. But, legally this notice isn't really a requirement anyway. Wikipedia seems to just suggest to use the year of initial publishing.

PR #99, opened 2018-08-24
rename prebuilt/ to bin/

mortend:
Before going Open Source, prebuilt/ used to be a place containing prebuilt
versions of various dependencies needed to build the solution. Built versions
of the uno-wrapper for loading and running the real uno-assemblies also ended
up in this directory.

The mentioned dependencies have since been removed and replaced by NuGet
packages, so we are now left with only the uno-wrapper. Therefore, I will
suggest to rename this directory to bin/, which is pretty much the industry
standard name for what this is. I think this will be more intuitive for
future users, and it's also shorter to type.

I don't really think updating a PATH variable, or similar, is a big enough
hassle to be a deal breaker, and right now I don't think we have very many
active developers anyway. Changes like this one don't happen often, and down
the line I think it will be worth it.


mortend:
The only external reference I've found touching this, is the Release Guide on our Wiki, section 5.2.2: https://github.com/fuse-open/uno/wiki/Release-Guide

PR #122, opened 2018-10-18
set Fuse Open as publisher of Uno packages

mortend:
This updates meta data information in our Uno projects.


mortend:
Since this is approved, I'll merge it now and start preparing the release branch.

PR #127, opened 2018-11-13
FuseJS.Transpiler: upgrade cached-path-relative

kusma:
Due to CVE-2018-16472, it's recommended to upgrade this package. As
the *only* change betweem 1.0.1 and 1.0.2 is the fix for this
vulnerability, let's upgrade it.

The vulnerability doesn't seem to be a big deal to our users, as they
control the scripts that run in their apps themselves. So there's no
reason to back-port this to older releases.

### DIFF
--- a/lib/FuseJS.Transpiler/src/package-lock.json
+++ b/lib/FuseJS.Transpiler/src/package-lock.json
@@ -258,9 +258,9 @@
       "integrity": "sha1-hZgoeOIbmOHGZCXgPQF0eI9Wnug="
     },
     "cached-path-relative": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/cached-path-relative/-/cached-path-relative-1.0.1.tgz",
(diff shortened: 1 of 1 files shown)


kusma:
@mortend: I know you dismissed this CVE, and I think that was the right move. But perhaps we should also just update on master, as the update is rather minimal?

kusma:
Note that I don't have a setup for actually testing if this fixes the problem or not, I just did "npm update" and reverted all seemingly irrelevant changes.

mortend:
Cool, though this doesn't fix the vulnerability in the minified version of the script that we use at run-time. The minified script can be updated by running `npm run build`, and this will probably make the update less minimal.

The minified script is only run locally on a developers computer in a dedicated `node` process per file when transpiling ES2015 files, and I don't think we expose any big risks even if we don't update.

kusma:
Right, makes sense. I guess it's better to just update everything and rebuild instead, then...

PR #130, opened 2018-11-20
Moved google repo higher

AndrewEQ:
Need to bump up google else fails to find certain dependencies:

https://stackoverflow.com/questions/50563407/could-not-find-play-services-basement-aar


mortend:
I ran into this problem too, but after a few days it went away again since it was a server-side issue at Google.

However, your patch doesn't do what you want since moving that XML element to the top doesn't actually change anything in the generated build script... This patch worked for me: https://github.com/mortend/uno/commit/9ccd1f238ceae49fa832fad1072e57240c1e20b8

AndrewEQ:
Ok cool, thanks.

mortend:
Closing this now.

Personally I haven't seen this problem except for that one time it happened to me, and the problem seems to be a hickup at Google. If it keeps happening in the future, we can open a new PR with the working patch.

PR #131, opened 2019-01-14
Add build target for metadata

mortend:
Example:

    uno build metadata --out-dir .uno/metadata

This will produce reference assemblies (one for each package) that contain
metadata suitable for code-completion and refactoring plugins, saved to the
specified location.


mortend:
The new `MetadataBackend` is a simplified/special cased version of `CilBackend`, producing reference assemblies that contain type information from the different Uno packages, but not actual runnable code. `MetadataBackend` also doesn't map Uno types to .NET types, which means the reference assemblies contain definitions for `Uno.Int`, `Uno.String` and so on. The idea is to use this metadata to create better code-completion plugins.

PR #136, opened 2019-01-14
C++: don't output source line numbers

mortend:
The source line numbers aren't really useful, and I think the output is
cleaner without.


mortend:
The generated C++ code contains comments like
```
// public override sealed string ToString() :123
void Foo__ToString_fn(Foo* __this, uString** __retval)
{
}
```

The `:123` at the end of the comment means the original method was declared on line 123, but that information isn't really useful and this patch removes it.

kusma:
@mortend: Right, thanks for clarifying! Makes sense to me!

PR #137, opened 2019-01-14
Native: disable warning on MSVC

mortend:
'op_Implicit' is a reserved word in the C++ backend and is output as
'op_Implicit1' in generated code. MSVC still isn't happy, so this patch
disables the following warning:

    warning C4674: 'op_Implicit' should be declared 'static' and have exactly one parameter


kusma:
I don't really understand what this warning is about or why we get the warning. But I also don't really like global warning disables, so I'm not the biggest fan... Is this the only way of dealing with this?

mortend:
We're already dealing with this by rewriting all occurrences of `op_Implicit` to `op_Implicit1`. Basically `op_Implicit`, `op_Explicit` and a bunch of other identifiers are treated as C++ keywords which means the compiler will automatically append numbers to turn them into non-keyword identifiers.

For some reason the mangled identifiers are still picked up by the MSVC compiler which gives warnings, which kind of seems like a bug in the MSVC compiler to me.

I don't know what the warning is about. I'm guessing they're trying to enforce a Microsoft specific pattern of some kind, but that seems unrelevant to us. It might be possible to mangle keywords in a different way to avoid the warning, but that also means mangling _all_ other keywords differently.

I encounter this warning sometimes, and instead of investigating more and since it doesn't look dangerous, I ended up disabling the warning to avoid the noise.

PR #139, opened 2019-01-14
Uno.Compiler: fix name resolver bug

mortend:
This fixes bugs when accessing members of a parent class from inside a nested
block.


mortend:
I noticed this bug while writing some new code that use `draw` statements.

The fix only applies to `draw` statements and meta property blocks and is very unlikely to break existing code. I don't think we need to notify users, other than making a note in CHANGELOG. I'll do that later.

PR #146, opened 2019-01-16
Run more tests on CI

mortend:
This fixes up test.sh and brings back UnoTest, UXTest and our old compiler
tests, which are currently not being run.

Both AppVeyor and Travis now run tests by invoking test.sh, which reduces
duplication and seems to be cleaner than it was before.


kusma:
"Exited with exception: Timed out while waiting for uno process to connect." 😢

mortend:
Seems to work after I skipped some tests when testing native on AppVeyor.

PR #147, opened 2019-01-18
NPM distribution

mortend:
This makes Uno runnable by Node.js and distributable via NPM.

Starting from version 1.11, Uno should be distributed standalone via npmjs.com, known as `@fuse-open/uno`. The plan is to also turn Fuselibs into a NPM package, `@fuse-open/fuselibs`, which will then specify a version of `@fuse-open/uno` as its dependency.

This means that we can easily install Uno and Fuselibs everywhere using `npm install`, that any project can control the exact version of Uno and Fuselibs they want to install, and that other NPM features such as `npm upgrade` and `npm link` will just work. This will also make things convenient when setting up CI building Uno code and Fuse apps, and we no longer need to wait for a new version of Fuse Studio to acquire new fixes in Uno or Fuselibs.

Disk space on our MyGet feed is running low and support for generating new NuGet packages has been dropped. Fuse Studio can get its next releases of Uno and Fuselibs from NPM, which I also think will simplify the release process and make it drastically easier to build Fuse Studio on custom versions of Uno and Fuselibs.


AndrewEQ:
Shweet :)

mortend:
I've published an NPM package that automatically sets up the Android SDK with required components at https://www.npmjs.com/package/android-build-tools

kusma:
OK, so this is pretty damn cool, and I think moving to using NPM in this way is totally the right move!

Unfortunately, it's a pretty big PR, and I don't have a lot of time to review right now. If someone else can review, that would be cool :)

mortend:
Thank you @kusma and good to know you approve the concept.

I have the new setup working locally with Fuselibs, Fuse Studio and an app, so the PR should be ready to merge. Since I don't know who else is around to review, I think it's fine to merge this myself. :)

PR #148, opened 2019-01-27
Drop Package Manager legacy

mortend:
This drops the following commands related to legacy package management.
```
  install        Install Uno package(s) to the local cache	
  uninstall      Uninstall Uno package(s) from the local cache	
  feed           Maintain Uno package feeds	
  pack           Build project(s) and make Uno package(s)	
  push           Upload Uno package(s) to remote server	
  stuff          Built-in simple package manager
```


mortend:
I have successfully used the new package manager on my local machine, by installing and using the package file generated by NPM. I also want these patches in the new release since they clean things up and shave off a few megabytes, and I might follow up with some more clean ups before I start on a new release.

The way things are now with so little activity and so few developers around, I don't think we should worry too much about process. I think it's more effective to merge now and possibly revert later if the new package manager turns out to be a failure, rather than spending energy doing test-deployments of things I already feel confident about...

kusma:
I think the lack of developers is exactly the reason why we shouldn't jump the gun and remove the code for the old solution; that way we can easily fall back to the previous solution in case of disaster...

kusma:
@mortend: Just FYI, I think the new package manager has proven itself by now, so if you want to rebase this, I wouldn't object :)

mortend:
Rebased now. :)

PR #153, opened 2019-03-09
Fix dependency downloads

mortend:
It looks like all dependencies hosted on files.fusetools.com are no longer available, effectively breaking all versions of Fuse and Fuse Studio ever released. Hopefully they will come back again soon...

Dependencies are now repackaged and hosted on nuget.org to get things up and running again and to avoid problems with files.fusetools.com in the future.

### DIFF
--- a/lib/UnoCore/Backends/CIL/CIL.uxl
+++ b/lib/UnoCore/Backends/CIL/CIL.uxl
@@ -31,8 +31,7 @@
     <Template Name="Uno.Native">
         <Require Assembly="@(PACKAGE_DIR)/prebuilt/Uno.Native/Uno.Native.dll" />
         <Require UnmanagedLibrary="@(PACKAGE_DIR)/prebuilt/Uno.Native/libuAsset.dylib" Condition="HOST_MAC" />
-        <Require UnmanagedLibrary.x86="@(PACKAGE_DIR)/prebuilt/Uno.Native/x86/uAsset.dll" Condition="HOST_WIN32" />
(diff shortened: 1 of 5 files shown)


lib/UnoCore/Backends/CIL/CIL.uxl
```diff
@@ -31,8 +31,7 @@
     <Template Name="Uno.Native">
         <Require Assembly="@(PACKAGE_DIR)/prebuilt/Uno.Native/Uno.Native.dll" />
         <Require UnmanagedLibrary="@(PACKAGE_DIR)/prebuilt/Uno.Native/libuAsset.dylib" Condition="HOST_MAC" />
-        <Require UnmanagedLibrary.x86="@(PACKAGE_DIR)/prebuilt/Uno.Native/x86/uAsset.dll" Condition="HOST_WIN32" />
-        <Require UnmanagedLibrary.x64="@(PACKAGE_DIR)/prebuilt/Uno.Native/x64/uAsset.dll" Condition="HOST_WIN32" />
+        <Require UnmanagedLibrary="@(PACKAGE_DIR)/prebuilt/Uno.Native/uAsset.dll" Condition="HOST_WIN32" />
```
kusma: Don't we need both x86 and x64?

mortend: Windows 32-bit support has been broken for a long time and it doesn't look like it's missed since no one have mentioned it. So I dropped the old x86 DLLs while repackaging to save some bandwidth/avoid some noise...

PR #155, opened 2019-03-22
FuseJS.Transpiler: upgrade cached-path-relative

kusma:
Due to CVE-2018-16472, it's recommended to upgrade this package. As
the *only* change betweem 1.0.1 and 1.0.2 is the fix for this
vulnerability, let's upgrade it.

The vulnerability doesn't seem to be a big deal to our users, as they
control the scripts that run in their apps themselves. So there's no
reason to back-port this to older releases.

### DIFF
--- a/lib/FuseJS.Transpiler/src/package-lock.json
+++ b/lib/FuseJS.Transpiler/src/package-lock.json
@@ -258,9 +258,9 @@
       "integrity": "sha1-hZgoeOIbmOHGZCXgPQF0eI9Wnug="
     },
     "cached-path-relative": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/cached-path-relative/-/cached-path-relative-1.0.1.tgz",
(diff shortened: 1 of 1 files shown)


kusma:
Here's some more details on the vulnerability: https://nvd.nist.gov/vuln/detail/CVE-2018-16472

mortend:
Identical to #127 - the problem was that we need to rebuild the minified script for changes to take effect.

Related to this, I also did some work on FuseJS.Transpiler recently: https://github.com/fuse-open/uno/compare/master...mortend:fusejs. My branch upgrades all dependencies to newest versions and adds support for ES2016..2018 and TypeScript while at it. I'll open a PR soon.

I'm also fine with merging this PR even if only to get a record about CVE-2018-16472 in our history.

kusma:
Doh, I forgot that I sent this in the past, and found it on some old branch. Sorry for the noise.

PR #156, opened 2019-03-22
UnoDoc: always support linking

kusma:
Without this, we'll fail generating the reference documentation, because
we have some API that only have C++ API (the corresponding .NET API is
provided by different, API-compatible means).


kusma:
I'm submitting this against 1.11, because it seems to be the only way of *cleanly* generating the 1.11 docs.

We might end up doing this manually, and in that case we might be fine with sending this against master instead.

mortend:
Do you think we should publish a v1.11.1 including this?

kusma:
@mortend: Nah, after hacking a bit more, I realized that more work is needed to make things clean (see newly created tickets), so I guess we could just put this on master instead.

PR #160, opened 2019-03-26
FuseJS: Add ES2016..2018 and TypeScript

mortend:
This upgrades to latest Babel 7 and adds support for ES2016, ES2017, ES2018 and TypeScript. The other dependencies are also upgraded, fixing the CVE-2018-16472 vulnerability #127 #155.

### TypeScript remarks

Note that Babel does not type-check its input. For that, you will need to install and set up TypeScript.

#### Caveats

- Does not support namespaces.
  Workaround: Move to using file exports, or migrate to using the module { } syntax instead.
- Does not support const enums because those require type information to compile.
  Workaround: Remove the const, which makes it available at runtime.
- Does not support export = and import =, because those cannot be compiled to ES.next.
  Workaround: Convert to using export default and export const, and import x, {y} from "z".

All these features are easy to avoid, or if necessary it's still possible to install and set up the official TypeScript compiler to run before running FuseJS.

For more information, see https://devblogs.microsoft.com/typescript/typescript-and-babel-7/


mortend:
We do have build-rules... but no automatic ways to invoke them during builds since we don't currently support build hooks in `uno`. Another thing is that it takes a long time to minify the JavaScript, longer than it takes to build all the other parts of `uno`, which is a little bit annoying when you're building `uno` on a regular basis... Since changes to the transpiler are pretty rare I think it's better to manually run minify when needed.

kusma:
OK, thanks for reminding me.

PR #171, opened 2019-04-17
Android: hardcode ${buildDir}

mortend:
We want this value to to match 'build' in https://github.com/fuse-open/uno/blob/master/lib/UnoCore/Targets/Android/app/src/main/CMakeLists.txt#L19,
but ${buildDir} don't expand to 'build' and instead gives '${buildDir}'...

Maybe something changed with quoting in newer gradle versions or maybe the
variable don't exist anymore, but hardcoding the value fixes the problem.


mortend:
Yeah, I'm sure this worked at one point...

PR #172, opened 2019-04-24
Stuff cleanup

mortend:
Mostly factored out of #148.

This refactoring moves/adapts useful parts of `Uno.Stuff` to `Uno.Build` and `Uno.Configuration`, and deletes the rest of `Uno.Stuff.dll`.

This PR also fixes two small bugs in the functionality:
1. Race condition in logging where sometimes the outputted color is not right (c517c2b).
2. (Handled) exceptions thrown during build (b91cc15).


kusma:
What's the difference between this and #148?

mortend:
This one doesn't remove the Uno Package Manager based on NuGet (1b00da9).

src/engine/Uno.Build/Stuff/Disk.cs
```diff
@@ -91,7 +91,10 @@ public static void TouchFile(Log log, string name)
         {
             try
             {
-                File.SetLastWriteTimeUtc(name, DateTime.UtcNow);
+                if (!File.Exists(name))
+                    File.Create(name);
+                else
+                    File.SetLastWriteTimeUtc(name, DateTime.UtcNow);
```
kusma: Technically speaking, I think this is still racy, as groups of filesystem operations aren't atomic. But I doubt it'll ever matter.

mortend: If two processes ends up here at the same time, one of the processes might throw but that's fine since the exception is ignored and the file should get it's timestamp updated by one of the processes, which is what we want.

mortend:
> Nice!
> 
> Just to make sure I understand it correctly. This PR:
> 
> 1. Turns stuff-file consumption into a core uno feature. This is needed for backwards compatibility.
> 2. Removes the possibility to produce new Stuff-files.
> 
> Right? If so, great!

Your understanding is accurate. Thanks for your review!

PR #173, opened 2019-04-24
AppLoader cleanup

mortend:
Factored out of #148.

This removes some unneeded parts of AppLoader related code.


mortend:
Slightly related, I think about setting an environment variable in Uno.TestRunner to tell AppLoader to hide the OpenGL window while running tests, since nothing interesting is happening inside the window...

kusma:
That sounds like a nice move.

In a fantasy world, I would really prefer if we didn't have all this apploader-magic, and rather exposed all the window creation code to uno (through foreign code, I suppose?), and just had a traditional entry-point like in C#. We could maintain backwards compatibility for end-users by providing that main function as part of FuseCore or something like that. But that's a lot of churn for little real-world gain.

mortend:
I agree and am hoping to get there eventually.

PR #175, opened 2019-04-25
Linux native build support

mortend:
This adds support for building native apps and running tests on Linux. Tested on Ubuntu 18.04 x86_64.


kusma:
We should probably also add the linux-build to travis-ci.

PR #177, opened 2019-04-30
Various UnoCore improvements

mortend:
Avoid a bunch of string allocations/copies on C++, verify some arguments, and add `GetString(byte[], int, int)` overloads for `Ascii` and `Utf8` to decode a subset of the array.

### DIFF
--- a/lib/UnoCore/Source/Uno/String.uno
+++ b/lib/UnoCore/Source/Uno/String.uno
@@ -470,6 +470,9 @@ namespace Uno
 
         public static string Join(string separator, params object[] value)
         {
+            if (value == null)
+                throw new ArgumentNullException(nameof(value));
+
             var strings = new string[value.Length];
 
             for (int i = 0; i < value.Length; i++)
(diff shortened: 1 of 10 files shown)


lib/UnoCore/Source/Uno/String.uno
```diff
@@ -482,17 +485,47 @@ namespace Uno
...
+                        memcpy(result->_ptr + offset, separator->_ptr, separator->_length * sizeof(@{char}));
+                        offset += separator->_length;
+                    }
+
+                    if (value->Unsafe<uString*>(i))
+                    {
+                        memcpy(result->_ptr + offset, value->Unsafe<uString*>(i)->_ptr, value->Unsafe<uString*>(i)->_length * sizeof(@{char}));
+                        offset += value->Unsafe<uString*>(i)->_length;
+                    }
+                }
```
kusma: Maybe assert that offset == length after the loop?

mortend: Good idea!

PR #178, opened 2019-04-30
Build for Linux on Travis

mortend:
This makes sure things build on Linux.

Because we get Segmentation Fault on Travis when we run tests, we only build them. Locally, tests run fine though.

### DIFF
--- a/lib/UnoCore/Source/Uno/IO/Directory.uno
+++ b/lib/UnoCore/Source/Uno/IO/Directory.uno
@@ -273,7 +273,7 @@ namespace Uno.IO
             @}
             else if defined(CPLUSPLUS)
             @{
-                chdir(uCString($0).Ptr);
+                (void)chdir(uCString($0).Ptr);
             @}
             else
                 throw new NotImplementedException();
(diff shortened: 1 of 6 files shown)


lib/UnoCore/Source/Uno/IO/Directory.uno
```diff
@@ -273,7 +273,7 @@ namespace Uno.IO
             @}
             else if defined(CPLUSPLUS)
             @{
-                chdir(uCString($0).Ptr);
+                int res = chdir(uCString($0).Ptr);
```
kusma: I think it's cleaner to just do `(void)chdir(uCString($0).Ptr);`

mortend: This will avoid an unused variable, so let's do that instead. Thanks!

mortend: Casting to `void` doesn't silence the warning... :/

https://travis-ci.org/fuse-open/uno/jobs/527629788#L2097

PR #179, opened 2019-05-02
Don't show window when running tests

mortend:
Since nothing is shown in the window when running tests it's better to hide
the window in this scenario.


mortend:
Thought I should say why the feature is there, but ended up dropping the comments. :)

PR #183, opened 2019-05-13
Mac and Windows icon properties

mortend:
This adds project properties to set icons for Mac and Windows apps.


lib/UnoCore/Backends/CPlusPlus/Uno/_main.cpp
```diff
@@ -55,6 +55,13 @@ struct uMainLoop : Xli::WindowEventHandler
 
         uBase::Auto<Xli::Window> wnd = Xli::Window::Create(uBase::Vector2i(375, 667), "@(Project.Title)", Xli::WindowFlagsResizeable);
 
+#if WIN32
+        HWND hWnd = (HWND)wnd->GetNativeHandle();
+        HICON hIcon = LoadIcon(GetModuleHandle(0), MAKEINTRESOURCE(101));
```
kusma: 101 seems a bit awkwardly hardcoded, it's common to have a resource.h that defines this mapping instead. But with only one icon, I guess this isn't too bad, and it's very unlikely to incrase. So, yeah... Let's go for this :)

PR #185, opened 2019-06-12
Apple Universal Link & Android App Link Support via Interapp

AndrewEQ:
https://developer.apple.com/ios/universal-links

https://developer.android.com/training/app-links

Todo: Add usage to Interapp docs

### DIFF
--- a/lib/UnoCore/Targets/Android/com/fuse/App.java
+++ b/lib/UnoCore/Targets/Android/com/fuse/App.java
@@ -217,6 +217,19 @@ public class App {
             pendingURI = intent.getDataString();
         }
 #endif
+
+#if @(Project.Android.AssociatedDomains:IsSet)
+        String[] associatedDomains = new String[]{@(Project.Android.AssociatedDomains:SplitAndJoin(',','"','"'))}; 
+
+        for (int i = 0; i < associatedDomains.length; i++) {
+
+            if (intent!=null && intent.getScheme()!=null && intent.getScheme().contains("http") 
+                && intent.getDataString().toLowerCase().contains(new String(associatedDomains[i]).toLowerCase())) { 
+                pendingURI = intent.getDataString();
+                break;
+            }
+        }
+#endif
         com.fuse.Activity._onActivityIntent(intent);
         @(Activity.OnNewIntent.Declaration:Join())
     }
--- a/lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
+++ b/lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
@@ -53,6 +53,16 @@
 #endif
                 @(AndroidManifest.Activity.ViewIntentFilter:Join('\n    ', '<data ', '/>'))
             </intent-filter>
+            
+#if @(Project.Android.AssociatedDomains:IsSet)
+            <intent-filter android:autoVerify="true">
+              <action android:name="android.intent.action.VIEW" />
+              <category android:name="android.intent.category.DEFAULT" />
+              <category android:name="android.intent.category.BROWSABLE" />
+              @(Project.Android.AssociatedDomains:SplitAndJoin('\n\t\t', '<data android:scheme="http" android:host="','" />'))
+              @(Project.Android.AssociatedDomains:SplitAndJoin('\n\t\t', '<data android:scheme="https" android:host="','" />'))
+            </intent-filter>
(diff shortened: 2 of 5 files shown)


lib/UnoCore/Targets/Android/com/fuse/App.java
```diff
@@ -134,7 +134,6 @@ public void onResume() {
         ActivityNativeEntryPoints.cppOnResume();
         com.fuse.Activity.onResume();
         focusOnResume();
-
```
kusma: This looks like an unneeded change. Please drop it.

lib/UnoCore/Targets/Android/com/fuse/App.java
```diff
@@ -217,6 +216,11 @@ public void onNewIntent (Intent intent)
             pendingURI = intent.getDataString();
         }
 #endif
+
+#if @(Project.Android.AssociatedDomains:IsSet)
+
+        @(Project.Android.AssociatedDomains:SplitAndJoin('\n\t\t', 'if (intent!=null && intent.getScheme()!=null && intent.getScheme().contains("http") && intent.getDataString().toLowerCase().contains(new String("','").toLowerCase())) { pendingURI = intent.getDataString(); }'))
```
kusma: Is there a reason for the SplitAndJoin-stuff here? Can't you just put the code in there verbatim instead?

lib/UnoCore/Targets/Android/com/fuse/App.java
AndrewEQ: Haha, I knew you would say that

lib/UnoCore/Targets/Android/com/fuse/App.java
AndrewEQ: Needed to cater for multiple domains in the .unoproj:
```
  "Android": {
    "AssociatedDomains": ["example.com","sub.example.com"],
   ...
```

lib/UnoCore/Targets/Android/com/fuse/App.java
AndrewEQ: Sorted.

lib/UnoCore/Targets/Android/com/fuse/App.java
AndrewEQ: Other syntax ideas for handling multiple?

kusma: How about generating an array of strings, and avoid repeating the code for each domain?

AndrewEQ: Oh yeah, I guess my mind got stuck in markup land... sorted.

lib/UnoCore/Targets/Android/com/fuse/App.java
```diff
@@ -208,6 +208,7 @@ public void onActivityResult (int arg0, int arg1, android.content.Intent arg2)
     //------------------------------------------------------------
 
     private String pendingURI = "";
+    private String[] associatedDomains;
```
kusma: I don't think you need a member-variable for this, a local variable should do just fine, no?

AndrewEQ: I guess so, just thought if someone else needed it in the future they could use it but sure, lemme localise it

AndrewEQ: Ok, sorted ;)

AndrewEQ:
Ha, thanks! 

In a nutshell, when someone taps on a web link in WhatsApp for example, it will open your app up and pass the url in, same behaviour for android.

kusma:
Ah, so more than just custom Uri schemes, like we already support? That's pretty cool :)

AndrewEQ:
Yeah, its more than the custom Uri scheme support; this is essentially binding the app to a domain that you have to verify, I routed the event through the existing InterApp modules receiving uri event to make things easier and well, code reuse.

So custom uri's will be like so: myapp://dosomething whereas, this is http://mydomain.com/dosomething

People don't exactly share urls like "myapp://dosomething" cos if the app is not installed, there's no fallback user experience, whereas with a url, it will just open up a browser. I think for android, if you used custom scheme to make a url, the difference is that android would then bring up a menu option asking you which app you would like to open the url with, whereas with "app links", they automatically just open the app 😄

PR #189, opened 2019-06-15
Android: use response file

mortend:
This will use a response file to pass arguments to linker, instead of the command-line, and enables us to use normal-length filenames on Android again. Because we now completely avoid the "command too long" issue on Windows.

### DIFF
--- a/lib/UnoCore/Targets/Android/app/src/main/CMakeLists.txt
+++ b/lib/UnoCore/Targets/Android/app/src/main/CMakeLists.txt
@@ -14,6 +14,13 @@ if (CMAKE_BUILD_TYPE STREQUAL "Release")
     set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -s")
 endif()
 
+# Using response file as workaround for long paths on Windows.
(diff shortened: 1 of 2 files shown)


kusma:
Hmm, are we sure these two fix the same problem? My understanding is that response files are about max argument length, but the truncation is about max path length... Those might interact at times, but won't max path length still matter?

mortend:
Max path length per file isn't an issue - they'll still be truncated by the backend if they get too long, but on Android we had to truncate more aggressively to work around bugs.

The current solution (before this patch) isn't robust - there's also a limit on number of source files we can build, because all the filenames are passed via one long command-line, and I think the maximum command length is 8000-something characters on Windows. I recently hit that limit in one of my projects  which led me to implement this solution. After setting up the response file I could also remove the aggressive truncating on Android.

We used response file in the past, in our pre-Gradle build-system, to fix the same problem. That code disappeared when we moved to Gradle and the solution became to make very short filenames instead. IIRC we've hit this wall before, and the solution last time was to make filenames even shorter. Response file seems to be a robust solution to the problem and allows us to revert the short filename workaround.

kusma:
Ah, I somehow confused this with the Windows max-path problem. Thanks for clearing that up!

PR #192, opened 2019-06-18
Native: always use Xcode on macOS

mortend:
Now we don't need to delete the CMake cache due to switching generators,
which speeds up subsequent builds.

Also, don't try to launch Xcode if passing --debug on Linux.


mortend:
We still use CMake on macOS, but now consistently use the Xcode generator in both normal builds and `--debug` builds, instead of `make` in the former scenario and Xcode in the latter.

PR #196, opened 2019-06-20
FuseJS: fix transpiled path

mortend:
GetFileNameWithoutExtension() strips directory components which gives us a
problem when transpiling files contained in a sub-directory.

Using ChangeExtension() instead fixes the problem.

### DIFF
--- a/src/engine/Uno.Build/Packages/PackageCache.cs
+++ b/src/engine/Uno.Build/Packages/PackageCache.cs
@@ -178,7 +178,7 @@ namespace Uno.Build.Packages
                     continue;
 
                 var name = f.NativePath;
-                var nameWithoutExt = Path.GetFileNameWithoutExtension(name);
+                var nameWithoutExt = Path.ChangeExtension(name, null);
                 var inputFile = Path.Combine(project.RootDirectory, name);
                 var outputFile = Path.Combine(project.RootDirectory, ".uno", "fusejs", nameWithoutExt + ".js");
 


src/engine/Uno.Build/Packages/PackageCache.cs
```diff
@@ -178,7 +178,7 @@ SourcePackage GetPackage(Source src, Project project, bool startup = false)
                     continue;
 
                 var name = f.NativePath;
-                var nameWithoutExt = Path.GetFileNameWithoutExtension(name);
+                var nameWithoutExt = Path.ChangeExtension(name, null);
```
kusma: Sounds like it fixes the problem it tries to.

mortend:
We change extension when transpiling `.ts` files to `.js` so `require()` in `Fuse.Scripting.JavaScript` can find them. If we keep the extension and end up with two `foo.js` and `foo.ts`, we'll have to solve the ambiguity in `Fuse.Scripting.JavaScript` instead, and teach `require()` about `.ts` files.

Our current solution is pretty simple and I think it works well for now. If ambiguity becomes a problem, I think a simple solution can be to print a warning to notify the user about the problem.

mortend:
But we're prone to ambiguity in other cases as well, for example two different Uno packages both containing a `foo.js`...

kusma:
Yeah, I'm aware that we have ambiguity-issues, but in this case I was afraid of two files being transpiled to the same source-name. It's of course still solvable by renaming the source files, so it's not a big deal either way.

PR #198, opened 2019-06-20
Logging tweaks

mortend:
Some subtle tweaks, mostly to some logging code.

### DIFF
--- a/src/compiler/Uno.Compiler.Core/BuildEnvironment.cs
+++ b/src/compiler/Uno.Compiler.Core/BuildEnvironment.cs
@@ -54,7 +54,7 @@ namespace Uno.Compiler.Core
             Extensions = extensions;
             Options = options;
             BundleDirectory = Path.Combine(project.CacheDirectory, "bundle");
-            CacheDirectory = Path.Combine(project.CacheDirectory, "builds", (options.BuildTarget + options.Configuration).GetHashCode().ToString("X8"));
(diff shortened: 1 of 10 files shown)


mortend:
This one should be good to go now...

src/compiler/Uno.Compiler.Core/BuildEnvironment.cs
```diff
@@ -54,7 +54,7 @@ public class BuildEnvironment : LogObject, IEnvironment
             Extensions = extensions;
             Options = options;
             BundleDirectory = Path.Combine(project.CacheDirectory, "bundle");
-            CacheDirectory = Path.Combine(project.CacheDirectory, "builds", (options.BuildTarget + options.Configuration).GetHashCode().ToString("X8"));
+            CacheDirectory = Path.Combine(project.CacheDirectory, "builds", (options.BuildTarget + options.Configuration).GetHashCode().ToString("x8"));
```
kusma: I know the commit message says this is backwards compatible, but how come it is?

mortend: The cache files are regenerated automatically when not found, so renaming the directory has no other consequences than improving naming consistency.

kusma: Right. So you'll end up with two cache-directories until the cache gets cleared?

mortend: Yes.

PR #208, opened 2019-07-25
pack.sh: support building outside of Git

mortend:
Don't terminate the script if a git command fails.

This fixes building .tar.gz files downloaded from GitHub, reported broken in
https://forums.fusetools.com/t/unable-to-compile-uno-fatal-not-a-git-repository/7817/1

### DIFF
--- a/scripts/pack.sh
+++ b/scripts/pack.sh
@@ -10,13 +10,16 @@ rm -rf ${DST:?}/* 2> /dev/null || :
 mkdir -p $DST
 
 # Detect version info
-COMMIT=`git rev-parse HEAD`
+COMMIT=`git rev-parse HEAD 2> /dev/null || :`
 VERSION=`cat package.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[\",]//g' | tr -d '[[:space:]]'`
 
 if [ -n "$APPVEYOR_REPO_BRANCH" ]; then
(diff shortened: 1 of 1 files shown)


scripts/pack.sh
```diff
@@ -10,13 +10,16 @@ rm -rf ${DST:?}/* 2> /dev/null || :
...
 
 if [ -n "$APPVEYOR_REPO_BRANCH" ]; then
     BRANCH=$APPVEYOR_REPO_BRANCH
-else
+elif [ -n "$COMMIT" ]; then
     BRANCH=`git rev-parse --abbrev-ref HEAD`
+else
+    BRANCH="unknown"
+    COMMIT="unknown"
 fi
```
kusma: Hmm, this all sounds a bit backwards to me.

Wouldn't it be better if we separated CI-magic in CI-scripts, and instead passed the branch name and commit id into scripts/pack.sh?

I mean, it's much cleaner if we have a set of scripts that are supposed to be called from the CI environment, and a set of scripts that doesn't care where it's called from...

mortend: Yeah, the branch/commit detection is a bit awkward, but replacing the mechanism will require some more work and we can probably improve things in a future PR.

This PR will as-is fix the immediate problem for users trying to build our source code archives at least...

kusma: Fair enough.

PR #209, opened 2019-07-25
Uno.Configuration: expand ~ to HomeDirectory

mortend:
Before, `~` expanded to `%HOME%` on all platforms which is not robost.

Now, we get `~` in a more robost way that provides several fallbacks and throws an exception if the directory could not be found.

This patch fixes so that more users will get their `~/.unoconfig` loaded properly.

### DIFF
--- a/src/common/Uno.Common/Diagnostics/PlatformDetection.cs
+++ b/src/common/Uno.Common/Diagnostics/PlatformDetection.cs
@@ -17,6 +17,59 @@ namespace Uno.Diagnostics
         public static readonly bool IsArm;
         public static readonly bool Is64Bit;
 
+        public static string HomeDirectory
+        {
+            get
+            {
+                var home = Environment.GetEnvironmentVariable("HOME");
+                if (!string.IsNullOrEmpty(home))
+                    return home;
+
+                if (IsWindows)
+                {
+                    var userProfile = Environment.GetEnvironmentVariable("USERPROFILE");
+                    if (!string.IsNullOrEmpty(userProfile))
+                        return userProfile;
+
+                    var homePath = Environment.GetEnvironmentVariable("HOMEDRIVE") +
+                                   Environment.GetEnvironmentVariable("HOMEPATH");
+                    if (!string.IsNullOrEmpty(homePath))
+                        return homePath;
+
+                    goto THROW;
+                }
+
+                var user = Environment.GetEnvironmentVariable("LOGNAME");
+                if (string.IsNullOrEmpty(user))
+                    user = Environment.GetEnvironmentVariable("USER");
+                if (string.IsNullOrEmpty(user))
+                    user = Environment.GetEnvironmentVariable("LNAME");
+                if (string.IsNullOrEmpty(user))
+                    user = Environment.GetEnvironmentVariable("USERNAME");
+
+                if (IsMac)
+                {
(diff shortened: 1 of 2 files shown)


mortend:
This patch is part of solving `ERROR: Path Android.SDK.Directory was not found in .unoconfig`, when NPM is installed in a non-standard location, reported by a user on Slack.

src/common/Uno.Common/Diagnostics/PlatformDetection.cs
```diff
@@ -17,6 +17,10 @@ public static class PlatformDetection
         public static readonly bool IsArm;
         public static readonly bool Is64Bit;
 
+        public static string UserDirectory => IsWindows
+                ? Environment.GetEnvironmentVariable("USERPROFILE")
```
kusma: I think this is questionable. IIRC, `$USERPROFILE` isn't always the right environment variable on Windows. In Git for Windows, we try `$HOME` first, then fall back to `$HOMEDRIVE/$HOMEPATH`, then finally fall back to `$USERPROFILE`.

Not consulting `$HOME` would be a regression for people who setup a custom home-dir outside of `C:\Users\`, for instance...

mortend: Great input, thanks!

I'll change the implementation to be like Git for Windows.

mortend: Fixed now.

src/common/Uno.Common/Diagnostics/PlatformDetection.cs
```diff
@@ -17,6 +17,65 @@ public static class PlatformDetection
...
+                    if (!string.IsNullOrEmpty(userProfile))
+                        return userProfile;
+
+                    var homePath = Environment.GetEnvironmentVariable("HOMEDRIVE") +
+                                   Environment.GetEnvironmentVariable("HOMEPATH");
+                    if (!string.IsNullOrEmpty(homePath))
+                        return homePath;
+
+                    if (!string.IsNullOrEmpty(home))
+                        return home;
```
kusma: I think it's better to always respect $HOME when it's set. It indicates that the user explicitly wants that folder to be treated as the home directory. In fact, not doing so would be a regression, I think.

mortend: Node.js believes otherwise and always respect $USERPROFILE on Windows (https://nodejs.org/api/os.html#os_os_homedir).

But you're probably right. We always respected $HOME before and we don't want regressions.

src/common/Uno.Common/Diagnostics/PlatformDetection.cs
```diff
@@ -17,6 +17,59 @@ public static class PlatformDetection
...
+                        return userProfile;
+
+                    var homePath = Environment.GetEnvironmentVariable("HOMEDRIVE") +
+                                   Environment.GetEnvironmentVariable("HOMEPATH");
+                    if (!string.IsNullOrEmpty(homePath))
+                        return homePath;
+
+                    goto THROW;
+                }
+
```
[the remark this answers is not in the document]
mortend: Great, that means we can drop some parts of the code. Updated again now. Thanks for the link.

PR #211, opened 2019-08-05
Solve crash apps with windowIsTranslucent

ckarmy:
Solve crash apps with windowIsTranslucent together with screenOrienta…tion="portrait" on Android 8 Oreo.

https://www.reddit.com/r/androiddev/comments/ap5zdm/dont_use_windowistranslucent_together_with/
https://forums.fusetools.com/t/problem-with-android-api-27-and-later/7430
https://stackoverflow.com/questions/48072438/java-lang-illegalstateexception-only-fullscreen-opaque-activities-can-request-o/50832408#50832408

### DIFF
--- a/lib/UnoCore/Targets/Android/app/src/main/res/values/styles.xml
+++ b/lib/UnoCore/Targets/Android/app/src/main/res/values/styles.xml
@@ -5,7 +5,7 @@
 #if !@(Project.Mobile.ShowStatusbar)
         <item name="android:windowFullscreen">true</item>
 #endif
-        <item name="android:windowIsTranslucent">true</item>
+        <item name="android:windowIsTranslucent">@(Project.Android.WindowIsTranslucent || @'false')</item>
         <item name="android:windowBackground">@android:color/transparent</item>
     </style>
 


Library/Core/UnoCore/Targets/Android/app/src/main/res/values/styles.xml
```diff
@@ -5,7 +5,11 @@
 #if !@(Project.Mobile.ShowStatusbar)
         <item name="android:windowFullscreen">true</item>
 #endif
-        <item name="android:windowIsTranslucent">true</item>
+#if !@(Project.Android.WindowIsTranslucent)
+        <item name="android:windowIsTranslucent">@(Project.Android.WindowIsTranslucent)</item>
+#else
+        <item name="android:windowIsTranslucent">false</item>
+#endif
```
kusma: Can't you just do this?
```xml
<item name="android:windowIsTranslucent">@(Project.Android.WindowIsTranslucent || @"false")</item>
```

ckarmy: Good idea! Done!

mortend:
Thank you for creating a fix!

@kusma Maybe this should be cherry-picked for 1.12.2 since the patch seems to fix a serious problem for some users?

@ckarmy The fix is made on an old version of Uno and we need to rebase to a newer version to fix the conflicts. Because the fix is small it is probably easiest to reset everything to latest master, apply the fix again and force-push your brach.

Create a backup of your branch. Just in case.
```
git branch backup
```
Reset to lastest Uno.
```
git fetch origin
git reset --hard origin/master
```
Apply the fix again, commit and force-push:
```
git commit -am"Solve crash apps with windowIsTranslucent together with screenOrienta…  …
…tion="portrait" on Android 8 Oreo"
git push --force
```

ckarmy:
> Thank you for creating a fix!
> 
> @kusma Maybe this should be cherry-picked for 1.12.2 since the patch seems to fix a serious problem for some users?
> 
> @ckarmy The fix is made on an old version of Uno and we need to rebase to a newer version to fix the conflicts. Because the fix is small it is probably easiest to reset everything to latest master, apply the fix again and force-push your brach.
> 
> Create a backup of your branch. Just in case.
> 
> ```
> git branch backup
> ```
> 
> Reset to latest Uno.
> 
> ```
> git fetch origin
> git reset --hard origin/master
> ```
> 
> Apply the fix again, commit and force-push:
> 
> ```
> git commit -am"Solve crash apps with windowIsTranslucent together with screenOrienta…  …
> …tion="portrait" on Android 8 Oreo"
> git push --force
> ```

I do that but nothing happen?
image

mortend:
Whoops, we forgot to use the right remote. My bad.

```
git remote add upstream https://github.com/fuse-open/uno.git
git fetch upstream
git reset --hard upstream/master
```

Then you can open `styles.xml` and paste in the fix again, before committing and force-pushing.

kusma:
@mortend: There's two things I don't like about that:
1. It changes the default. This generally should make it unsuitable for back-porting as-is IMO, as it's not a pure bug-fix.
2. The fix might be questionable. The default-translucency is there for a reason; without it we get some problematic rendering-glitches when hiding and showing the keyboard for non-native applications. Without that fixed in some other way, I'm a bit reluctant about this fix even for master...

ckarmy:
@mortend Done! I'ts all OK?

Library/Core/UnoCore/Targets/Android/app/src/main/res/values/styles.xml
ckarmy: It's all done. But still showing `Changes requested`. i need to do anithing more?

Thanks!

kusma:
As for the "questionableness" of the fix; I think the translucent stuff has caused a lot of issues, and the glitches are probably better than crashes, hence merged ;)

PR #220, opened 2019-08-25
Android build Enhancements

ichan-mb:
#### Android build target
- Upgraded to Gradle 5.6, and Gradle-plugin 3.4.2.
- Upgraded build tools to version 28.0.3.
- Upgraded support libraries to version 28.0.0.
- Upgraded SDK compile and target versions to 28
- Added support for generating [Android App Bundle](https://developer.android.com/platform/technology/app-bundle) on Release
- Added the following build properties.
    * `Bundle`
    * `Bundle.BuildName`
    * `Bundle.Gradle.Task`
- Renamed the following build properties.
    * `APK.Configuration -> Build.Configuration`
- Set default versionCode to positive Integer (1) so `gradlew` command doesn't produce error when building using latest Gradle


kusma:
It's a bit hard to review this when all those changes are the same commit. Would you mind splitting it up into a commit per bullet point instead?

ichan-mb:
Okay, coming right up

lib/UnoCore/Targets/Android/build.sh
```diff
@@ -22,8 +22,14 @@ exit 1
 export JAVA_HOME="@(JDK.Directory)"
 #endif
 
-./gradlew @(Gradle.Task) "$@"
+./gradlew @(APK.Gradle.Task) "$@"
```
mortend: We also need similar changes in `build.bat` for Windows.

ichan-mb: Omg! I always forgot Windows users, I am sorry. fix coming right way

ichan-mb:
@kusma yes I think it should. fixed!

Issue #210, opened 2019-07-29
Testing - macOS Catalina - iOS13 - Run Fail


AndrewEQ:
Hey there,

`uno build -t=ios --run` successfully deployed the app but didn't run it, the error message returned after the build and deploy was:

```
Error 0xe800007f: Device doesn't support wireless sync.
```

ichan-mb:
I think the problem is on `ios-deploy` command. I see an issue has been raised in `ios-deploy` repo [here](https://github.com/ios-control/ios-deploy/pull/385) and has been merged. maybe we can use the latest version of `ios-deploy`

just google up, Flutter has the same issue as describe [here](https://github.com/flutter/flutter/issues/33909)  and replacing `ios-deploy` cmd with this one `wget https://raw.githubusercontent.com/kangwang1988/kangwang1988.github.io/master/others/ios-deploy` resolved the issue

mortend:
Good catch!

We should probably switch to `ios-deploy` (instead of `"@(Base.Directory)/bin/ios-deploy"`) in https://github.com/fuse-open/uno/blob/master/lib/UnoCore/Targets/iOS/run.sh.

If https://www.npmjs.com/package/ios-deploy includes the fix, we can probably add `ios-deploy` to `dependencies` in `package.json`, and let npm take care of installing things.

mortend:
The beta version on NPM should include the fix - `npm install ios-deploy@beta --save`.

ichan-mb:
Yeah, sounds a good idea!

mortend:
Resolved by #226. Closing issue.

AndrewEQ:
@mortend just tried it out again with latest Uno, and it doesn't seem to have permission to run the script, I'm using the compiled version of Uno:
```
run.sh: line 27: /Users/andrewq/www/github/fuse-open/uno/lib/UnoCore/prebuilt/iOS/bin/ios-deploy: Permission denied
```

mortend:
That's weird, the latest version of Uno runs `ios-deploy` from PATH and there are no longer any references to `[...]/lib/UnoCore/prebuilt/iOS/bin/ios-deploy` in our source code. Also, `ios-deploy` is executed on line 37 and 39 (not 27) in the latest version of the script, so it looks like you're not using the latest version.

[...]

PR #226, opened 2019-09-08
iOS: switch to user-installed ios-deploy

mortend:
Our version of `ios-deploy` that is included in `uno-base` is outdated and no longer works with the most recent version of iOS (#210).

To get around this problem we'll use the system-installed version instead.

Currently, the most recent beta version of `ios-deploy` is required for iOS 13, which can be installed by running this command.

```
npm install ios-deploy@beta -g
```

### DIFF
--- a/lib/UnoCore/Targets/iOS/run.sh
+++ b/lib/UnoCore/Targets/iOS/run.sh
@@ -17,17 +17,14 @@ debug)
     ;;
 uninstall)
     echo "Uninstalling @(BundleIdentifier)"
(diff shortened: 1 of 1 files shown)


kusma:
Don't we have to update the README about this added dependency?

mortend:
It is not a build requirement for Uno, but `ios-deploy` is needed as soon as you want to deploy your app to an iOS device.

I plan to add this information to the `fuse-sdk` package before we launch the next minor version. I should probably also move the `fuse-sdk` repo from my account to the `fuse-open` organization on GitHub.

PR #227, opened 2019-09-08
Detect required macOS software

mortend:
Provide some friendly hints when required software is missing on the system.

### DIFF
--- a/lib/UnoCore/Targets/Native/build.sh
+++ b/lib/UnoCore/Targets/Native/build.sh
@@ -6,6 +6,10 @@ cd "`dirname "$0"`"
 
 if ! which @(CMake) > /dev/null 2>&1; then
     echo "ERROR: Unable to find the 'cmake' command. Make sure CMake is installed and added to PATH." >&2
+#if @(MAC:Defined)
+    echo -e "\nOn macOS, you can install CMake using Homebrew:" >&2
+    echo -e "\n    brew install cmake\n" >&2
+#endif
     exit 1
 fi
 
--- a/lib/UnoCore/Targets/Native/build.sh
+++ b/lib/UnoCore/Targets/Native/build.sh
@@ -14,6 +14,13 @@ if ! which @(CMake) > /dev/null 2>&1; then
 fi
 
 #if @(MAC:Defined)
+if ! which xcodebuild > /dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


lib/UnoCore/Targets/Native/build.sh
```diff
@@ -6,10 +6,21 @@ cd "`dirname "$0"`"
 
 if ! which @(CMake) > /dev/null 2>&1; then
     echo "ERROR: Unable to find the 'cmake' command. Make sure CMake is installed and added to PATH." >&2
+#if @(MAC:Defined)
+    echo -e "\nOn macOS, you can install CMake using Homebrew:" >&2
+    echo -e "\n    brew install cmake\n" >&2
+#endif
```
kusma: Just a thought, should we mention the same thing about chocolatey for Windows? If so, we can of course do that separately, no need to update this PR for it.

lib/UnoCore/Targets/iOS/build.sh
```diff
@@ -5,6 +5,22 @@ set -e
 set -o pipefail
 cd "`dirname "$0"`"
 
+if ! which xcodebuild > /dev/null 2>&1; then
+    echo "ERROR: Unable to find the 'xcodebuild' command." >&2
+    echo -e "\nYou can download Xcode Command Line Tools from this page:" >&2
+    echo -e "\n    https://developer.apple.com/xcode/\n" >&2
+    exit 1
+fi
```
kusma: This one looks very familiar ;)

Is there any reasonable way we could define this as a function in some helper-source rather than repeating it? If not, twice isn't *too* bad, I just don't want us to have to remember to update two code-sites if a link changes or something...

lib/UnoCore/Targets/Native/build.sh
mortend: Good though. We should, but currently `@(CMake)` is configurable and the `where` command on Windows doesn't work well with full paths. I don't think `@(CMake)` need to be configurable and if we replace it with just `cmake`, the install-hint will work well on Windows too.

lib/UnoCore/Targets/iOS/build.sh
mortend: One is for iOS and the other for macOS. It is difficult to share this code between the two, so in this case I think code duplication is better than a bunch of engineering.

PR #228, opened 2019-09-08
macOS: install Mono automatically

mortend:
The [dotnet-run](https://www.npmjs.com/package/dotnet-run) package will detect Mono/.NET or automatically download a suitable version for your system.

> When we can find Mono v5.4.1.7 or greater on your system, we'll use your existing installation. By default we'll download v6.0.0.311 automatically.
>
> We provide minimal Mono releases for macOS, much smaller than official releases. Our v6.0.0.311 is distributed as a 33.6 MB tarball, which installs faster and is more than ten times smaller than the official pkg installer at 362.2 MB. Our package is unobtrusive to your system and will safely co-exist with any existing Mono installation you may or may not have from before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,7 @@ Welcome to Uno, the core component in [Fuse Open], a native app development tool
 npm install @fuse-open/uno
 ```
 
-This will install `uno`. Pass `--save-dev` to install as a dependency in your local project, or `-g` to install
-the command globally on your system.
-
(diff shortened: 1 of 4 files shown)


mortend:
The `dotnet-run` package is basically an updated and NPM-ified version of our old MonoGenerator tool in Fuse Studio, located at https://github.com/fuse-open/fuse-studio/tree/master/Tools/MonoGenerator.

PR #230, opened 2019-09-08
Config-file clean-ups

mortend:
Some refactoring on config files and related code.


config/common.unoconfig
```diff
@@ -6,6 +6,3 @@ if WIN32 {
     require "%HOME%/Library/Application Support/Fusetools/Fuse/Android/.sdkconfig" // For legacy 'fuse install android'
     require /usr/local/lib/node_modules/android-build-tools/.unoconfig // For 'npm install -g android-build-tools'
 }
-
-// User config
-require ~/.unoconfig
```
kusma: Nit: I would kinda have expected this to be squashed in with 70f322d...

mortend: Makes sense, I'll do that. :)

PR #235, opened 2019-09-18
Added iOS13 check for Sign In With Apple capability

AndrewEQ:
Added iOS13 check for SignInWithApple capability

### DIFF
--- a/lib/UnoCore/Targets/iOS/@(Project.Name)/@(Project.Name).entitlements
+++ b/lib/UnoCore/Targets/iOS/@(Project.Name)/@(Project.Name).entitlements
@@ -3,11 +3,13 @@
 <plist version="1.0">
 <dict>
 
+#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
 #if @(Project.iOS.SystemCapabilities.SignInWithApple:IsSet:Test(@(Project.iOS.SystemCapabilities.SignInWithApple),0))
 	<key>com.apple.developer.applesignin</key>
 	<array>
 		<string>Default</string>
 	</array>
+#endif
 #endif
 
 #if @(@(Project.iOS.SystemCapabilities.Push:IsSet):Test(@(Project.iOS.SystemCapabilities.Push),0))


lib/UnoCore/Targets/iOS/@(Project.Name)/@(Project.Name).entitlements
```diff
@@ -3,12 +3,14 @@
 <plist version="1.0">
 <dict>
 
+#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
```
mortend: `__IPHONE_13_0` and `__IPHONE_OS_VERSION_MAX_ALLOWED` are C preprocessor macros and can't be used inside XML files.

AndrewEQ: darn, its for the case of me handing you or a fellow dev the same project with the entitlement set but you don't have access to the entitlement, so it breaks when you try to build

mortend: Oh... That sounds like an iffy situation, but then plain native Xcode projects must have the same issue as well? How is this usually solved in Xcode projects?

With Uno, it should be possible to explicitly disable the feature from the command line, though, like so: 
```
uno build ios --set:Project.iOS.SystemCapabilities.SignInWithApple=0
```

Or, the other way around, explicitly enabling the feature when it's needed:
```
uno build ios --set:Project.iOS.SystemCapabilities.SignInWithApple=1
```

AndrewEQ: Yeah, unfortunately the way forward is to literally buy a new Mac, upgrade to new OS and Xcode lol, thanks @mortend, will give it a try

mortend: Yeah, that's Apple Style. Cheers @AndrewEQ!

AndrewEQ: I can confirm that `uno build ios --set:Project.iOS.SystemCapabilities.SignInWithApple=0` works! 😄 🎉  So now we just gotta remember to write about this somewhere 😆

PR #238, opened 2019-09-30
Simplify version info

mortend:
This drops $BRANCH and $SUFFIX, and simplifies logic (#208).

When updating an NPM package, the version field must be updated in
package.json, and all changes to package.json should be committed in Git.

This means that publishing an NPM package, pre-release or not, will always
involve updating package.json, committing, and pushing a tag in Git.

This also means that pack.sh should not manipulate version info read from
package.json, as it used to do in pre-NPM times for legacy reasons.

### DIFF
--- a/scripts/pack.sh
+++ b/scripts/pack.sh
@@ -10,36 +10,16 @@ rm -rf ${DST:?}/* 2> /dev/null || :
 mkdir -p $DST
 
 # Detect version info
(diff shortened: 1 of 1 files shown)


mortend:
Updated code and description.

mortend:
Merging this now because it will fix version info in upcoming canary builds. I plan to release the first canary build shortly...

PR #240, opened 2019-10-04
Android Splash

AndrewEQ:
- fixed flash of a screen when no splash
- added .unoproj option to add splash screen:

```
"Android": {
  "Splash": {
    "Enabled": true,
    "SplashBackgroundColor": "#9846EF",
    "SplashLogo": "Assets/splash_logo.png",		
  },
}
```

### DIFF
--- a/lib/UnoCore/Targets/Android/Android.uxl
+++ b/lib/UnoCore/Targets/Android/Android.uxl
@@ -226,6 +226,9 @@
     <ProcessFile Name="app/build.gradle" />
     <ProcessFile Name="app/src/main/AndroidManifest.xml" />
     <ProcessFile Name="app/src/main/CMakeLists.txt" />
+    <CopyFile Condition="Android" Name="@(Project.Android.Splash.SplashLogo:Path || '@//Assets/DefaultIcon.png')" TargetName="app/src/main/res/drawable/splash_logo.png" />
+    <ProcessFile Name="app/src/main/res/drawable/splash_background.xml" />
+    <ProcessFile Name="app/src/main/res/values/colors.xml" />
     <ProcessFile Name="app/src/main/res/values/strings.xml" />
     <ProcessFile Name="app/src/main/res/values/styles.xml" />
     <ProcessFile Name="build.gradle" />
--- a/lib/UnoCore/Targets/Android/app/src/main/res/drawable/splash_background.xml
+++ b/lib/UnoCore/Targets/Android/app/src/main/res/drawable/splash_background.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+ 
+ <item android:drawable="@color/splashColor" />
+ 
+ <item>
+	 <bitmap
+	 android:gravity="center"
+	 android:src="@drawable/splash_logo" />
+ </item>
+ 
+</layer-list>
--- a/lib/UnoCore/Targets/Android/app/src/main/res/values/styles.xml
+++ b/lib/UnoCore/Targets/Android/app/src/main/res/values/styles.xml
@@ -5,8 +5,16 @@
 #if !@(Project.Mobile.ShowStatusbar)
         <item name="android:windowFullscreen">true</item>
 #endif
-        <item name="android:windowIsTranslucent">@(Project.Android.WindowIsTranslucent || 'false')</item>
-        <item name="android:windowBackground">@android:color/transparent</item>
+
+#if @(Project.Android.Splash.Enabled:Test(1, 0))
+		<item name="android:windowBackground">@drawable/@(Project.Android.Splash.SplashFileName || 'splash_background')</item>
+#else
+		<item name="android:windowFullscreen">true</item>
+		<item name="android:windowContentOverlay">@null</item>
+		<item name="android:windowIsTranslucent">@(Project.Android.WindowIsTranslucent || 'true')</item>
+		<item name="android:windowBackground">@android:color/transparent</item>
+#endif
+        
     </style>
 
 </resources>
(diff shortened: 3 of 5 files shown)


lib/UnoCore/Targets/Android/app/src/main/res/values/styles.xml
```diff
@@ -5,8 +5,16 @@
...
-        <item name="android:windowIsTranslucent">@(Project.Android.WindowIsTranslucent || 'false')</item>
-        <item name="android:windowBackground">@android:color/transparent</item>
+
+#if @(Project.Android.Splash.Enabled)
+	<item name="android:windowBackground">@drawable/@(Project.Android.Splash.SplashFileName || 'splash_background')</item>
+#else
+	<item name="android:windowFullscreen">true</item>
+	<item name="android:windowContentOverlay">@null</item>
+	<item name="android:windowIsTranslucent">@(Project.Android.WindowIsTranslucent || 'true')</item>
+	<item name="android:windowBackground">@android:color/transparent</item>
```
mortend: Should indent `<item>`s with 6 spaces to align with other `<item>`s in this file.

lib/UnoCore/Targets/Android/com/fuse/App.java
```diff
@@ -88,6 +89,11 @@ public void onCreate(Bundle savedInstanceState)
 
         // call c++ for setup
         ActivityNativeEntryPoints.cppOnCreate(RootActivity);
+        
+        // reset window background for app switcher to be able to take a screenshot of app content
+        #if @(Project.Android.Splash.Enabled:IsSet)
+        RootActivity.getWindow().setBackgroundDrawableResource(android.R.color.transparent);
+        #endif
```
mortend: I personally prefer `#` directives to not be indented, but that's not a big deal.

lib/UnoCore/Targets/Android/Android.uxl
```diff
@@ -226,6 +226,9 @@
     <ProcessFile Name="app/build.gradle" />
     <ProcessFile Name="app/src/main/AndroidManifest.xml" />
     <ProcessFile Name="app/src/main/CMakeLists.txt" />
+    <CopyFile Condition="Android" Name="@(Project.Android.Splash.SplashLogo:Path || '@//Assets/DefaultIcon.png')" TargetName="app/src/main/res/drawable/splash_logo.png" />
```
mortend: Should `splash_logo.png` be in specific dimensions? If so, maybe we can use `<ImageFile>` instead, and pass `TargetWidth` and `TargetHeight`.

lib/UnoCore/Targets/Android/com/fuse/App.java
AndrewEQ: Quick question, how do I check if that project variable has been set and that its "true"?

AndrewEQ: Or maybe I should remove the "Enabled" property and just check if Android.Splash is set?

mortend: I think you can `#if @(Project.Android.Splash.Enabled:Test(1, 0)`. That'll default to `0` when not set.

mortend: You'll probably need the `Enabled` part, I think, or you'll have to check if any of `SplashBackgroundColor` or `SplashLogo` are set.

AndrewEQ: Ok, lemme try that; I've got defaults for the background colour and logo

lib/UnoCore/Targets/Android/Android.uxl
AndrewEQ: I think its upto the designer to make their logo larger or smaller no?

AndrewEQ: Would you prefer project properties for the width and height?

mortend: Hmm, instead of project properties, it's probably better to use the logo as-is, yeah. Maybe the log will be automatically centered/resized by Android if dimensions aren't right?

AndrewEQ: Yeah, we could put more constraints on the image here: https://github.com/fuse-open/uno/blob/4523619fc786cdd19c2818fd3075e215fdf61cb9/lib/UnoCore/Targets/Android/app/src/main/res/drawable/splash_background.xml

Just don't how if there is such a thing.

mortend: I see, yeah, it's probably fine as-is. I think we can add more constraints and expose those as project properties, if we later need more control over that stuff.

AndrewEQ: Alrighty

lib/UnoCore/Targets/Android/app/src/main/res/values/styles.xml
```diff
@@ -5,8 +5,16 @@
...
-        <item name="android:windowIsTranslucent">@(Project.Android.WindowIsTranslucent || 'false')</item>
-        <item name="android:windowBackground">@android:color/transparent</item>
+
+#if @(Project.Android.Splash.Enabled:Test(1, 0))
+		<item name="android:windowBackground">@drawable/@(Project.Android.Splash.SplashFileName || 'splash_background')</item>
+#else
+		<item name="android:windowFullscreen">true</item>
+		<item name="android:windowContentOverlay">@null</item>
+		<item name="android:windowIsTranslucent">@(Project.Android.WindowIsTranslucent || 'true')</item>
+		<item name="android:windowBackground">@android:color/transparent</item>
```
mortend: Nitpicking after taking a closer look: Indentation in this file should be 8 spaces and not two tabs.

lib/UnoCore/Targets/Android/app/src/main/res/drawable/splash_background.xml
```diff
@@ -0,0 +1,13 @@
...
+
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+ 
+ <item android:drawable="@color/splashColor" />
+ 
+ <item>
+	 <bitmap
+	 android:gravity="center"
+	 android:src="@drawable/splash_logo" />
+ </item>
```
mortend: It would be nice to be consistent about indentation across different XML-files, 4 spaces per level, and indent attributes one level more than their parent element.

PR #251, opened 2019-11-05
UnoCore: strip Byte Order Mark for UTF-8

mortend:
When loading files using BundleFile.ReadAllText().
This is consistent with File.ReadAllText().


kusma:
Maybe we should have a test for this also?

PR #252, opened 2019-11-06
Android - Added Dynamic Delivery options

AndrewEQ:
```
"Android": {
    "SupportedLanguages": ["en"],
    "Bundle": {
      "CompressNativeLibraries": true,
      "Language": {
        "EnableSplit": false
      },
      "Density": {
        "EnableSplit": true
      },
      "ABI": {
        "EnableSplit": true
      }
    },
    ...
```

Native Library Compression
- turn on/off compression: https://developer.android.com/studio/projects/dynamic-delivery/configure-base#base_manifest

App Supported Languages
- define apps official languages so that other language resources in plugins will be removed except for the apps official languages: https://developer.android.com/studio/build/shrink-code#unused-alt-resources

Bundle Base Split Options
- language, density and abi options are made available to turn on/off: https://developer.android.com/studio/projects/dynamic-delivery/configure-base#disable_config_apks

Supports Screens
Define which size screens your app supports with these properties: https://developer.android.com/guide/topics/manifest/supports-screens-element

```
"Android": {
"SupportsScreens" : {
"Resizable" : "true",
"NormalScreens" : "true",
"LargeScreens" : "true",
"XlargeScreens" : "true",
"AnyDensity" : "true",
"RequiresSmallestWidthDp": "320",
...
},
...
```

### DIFF
--- a/lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
+++ b/lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
@@ -15,7 +15,11 @@
                  android:description="@string/description"
                  android:largeHeap="@(Project.Android.LargeHeap:IsSet:Test(@(Project.Android.LargeHeap:Bool),false))"
                  android:allowBackup="@(Project.Android.AllowBackup:IsSet:Test(@(Project.Android.AllowBackup:Bool),false))"
-                 android:usesCleartextTraffic="@(Project.Android.UsesCleartextTraffic:IsSet:Test(@(Project.Android.UsesCleartextTraffic:Bool),false))">
+                 android:usesCleartextTraffic="@(Project.Android.UsesCleartextTraffic:IsSet:Test(@(Project.Android.UsesCleartextTraffic:Bool),false))"
+#if @(Project.Android.Bundle.CompressNativeLibraries:Test(1, 0))
+                 android:extractNativeLibs="false"
+#endif
+                 >
 
         <activity android:name="@(Activity.Name:EscapeXml)"
                   android:label="@(Project.Android.ApplicationLabel:EscapeXml)"
--- a/lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
+++ b/lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
@@ -2,6 +2,38 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="@(Activity.Package)">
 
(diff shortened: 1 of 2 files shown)


lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
```diff
@@ -15,7 +47,11 @@
                  android:description="@string/description"
                  android:largeHeap="@(Project.Android.LargeHeap:IsSet:Test(@(Project.Android.LargeHeap:Bool),false))"
                  android:allowBackup="@(Project.Android.AllowBackup:IsSet:Test(@(Project.Android.AllowBackup:Bool),false))"
-                 android:usesCleartextTraffic="@(Project.Android.UsesCleartextTraffic:IsSet:Test(@(Project.Android.UsesCleartextTraffic:Bool),false))">
+                 android:usesCleartextTraffic="@(Project.Android.UsesCleartextTraffic:IsSet:Test(@(Project.Android.UsesCleartextTraffic:Bool),false))"
+#if @(Project.Android.Bundle.CompressNativeLibraries:Test(1, 0))
+                 android:extractNativeLibs="false"
+#endif
```
mortend: I feel this project property should've been named `Android.ExtractNativeLibs`, similar as in AndroidManifest, maybe to make it more familar for Android users.

The whole `#if..#endif` block can probably be replaced with `android:extractNativeLibs="@(Project.Android.ExtractNativeLibs:IsSet:Test(@(Project.Android.ExtractNativeLibs:Bool),true))"`

lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,38 @@
...
+                  #if @(Project.Android.SupportsScreens.SmallScreens:IsSet)
+                  android:smallScreens="@(Project.Android.SupportsScreens.SmallScreens:ToLower)"
+                  #endif
+                  #if @(Project.Android.SupportsScreens.NormalScreens:IsSet)
+                  android:normalScreens="@(Project.Android.SupportsScreens.NormalScreens:ToLower)"
+                  #endif
+                  #if @(Project.Android.SupportsScreens.LargeScreens:IsSet)
+                  android:largeScreens="@(Project.Android.SupportsScreens.LargeScreens:ToLower)"
+                  #endif
+                  #if @(Project.Android.SupportsScreens.XlargeScreens:IsSet)
```
mortend: Maybe `XlargeScreens` would look better if it was `XLargeScreens`? That feels a little bit more natural to me at least.

Though, project property names are case-insensitive so either of those will compile...

lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
AndrewEQ: ok, done

lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
AndrewEQ: Thanks for spotting it, done

PR #254, opened 2019-11-11
AndroidX Support

AndrewEQ:
Migrated support libraries to AndroidX
https://developer.android.com/jetpack/androidx/migrate

### DIFF
--- a/lib/Uno.Permissions/Permissions.java
+++ b/lib/Uno.Permissions/Permissions.java
@@ -1,9 +1,9 @@
 package com.fuse;
 
 import android.content.pm.PackageManager;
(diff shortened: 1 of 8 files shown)


mortend:
Very cool!

I think we can merge this as soon as we've started publishing canary-builds to NPM. Then we can upgrade to a compatible Uno version in the corresponding Fuselibs PR, to make all builds pass.

if this breaks any third-party libs, I suppose we'll just have to upgrade/fix things as we go. I think it's more important to stay up-to-date with the Android ecosystem.

AndrewEQ:
Sure thing boss 😄 👍

PR #255, opened 2019-11-11
Android - Android Go Support

AndrewEQ:
- added declaration for app optimized for Android Go: https://developer.android.com/docs/quality-guidelines/building-for-billions-device-capacity#androidgo

### DIFF
--- a/lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
+++ b/lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
@@ -34,7 +34,7 @@
                   />
     #endif
     
-#if @(Project.Android.SupportsAndroidGo:IsSet)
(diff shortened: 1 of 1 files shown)


lib/UnoCore/Targets/Android/app/src/main/AndroidManifest.xml
```diff
@@ -33,7 +33,10 @@
                   #endif
                   />
     #endif
-
+    
+#if @(Project.Android.SupportsAndroidGo:IsSet)
```
mortend: I think this line should be the following, so that the value of the property is respected. Otherwise it looks like we'll insert the same block whether the value is `true` or `false`.

```suggestion
#if @(Project.Android.SupportsAndroidGo:Test(1, 0))
```

PR #257, opened 2019-11-21
Auto-sync version info

mortend:
This adds a script that's automatically run when using `npm version`. This makes sure version information in the NPM package and .NET assemblies are in sync.


mortend:
It's very useful to get this in before making a new canary release.

PR #258, opened 2019-11-21
Improve Android AAR support

mortend:
This makes build-scripts aware of AAR (Android library) that are produced instead of APK when passing `-DLIBRARY`, and other fixes related to this.


mortend:
It's nice to get this in before making a new canary release. Dependents are waiting for a clean NPM way to install this, and it's tested locally...

PR #261, opened 2019-11-25
UnoCore: add missing file to project

mortend:
This adds a file reference that was missing in the project, thus missing from `npm pack`-ed builds, causing Android builds to fail since #254.


mortend:
This fixes a regression and is needed for the next canary release.

PR #292, opened 2020-03-11
Launch screen storyboard on iOS

ichan-mb:
Xcode project generated by Uno now using `LaunchScreen.storyboard` instead of LaunchImages to comply with new rules from Apple. More info: https://developer.apple.com/news/?id=03042020b

We also bump minimal ios version to ver 9.0


AndrewEQ:
It needs to reference a separate launch image

AndrewEQ:
Sorry, busy testing it, will send patch to support "dark mode launch" too.

AndrewEQ:
...just busy working out centering and scaling of an image, you'd think it would be simple... anyone got any ideas for this? 😄

PR #299, opened 2020-03-14
Updated iOS LaunchScreen

AndrewEQ:
- fixed ```Mobile.DisableStatusbar```.
- added ```Project.iOS.UIStatusBarHidden``` to set if the StatusBar is hidden on launch, false by default.
- supports dark mode dynamic background color, white for legacy devices.
- logo type of storyboard, supports a single image.
  - should make your logo or image support atleast x3 resolution.
- removed launch screen images, so even if you have them specified, they won't be added to the Xcode project.
- added .unoproj options:
  - all of the following options are optional.
  - if you don't specify the image, this will be used: ```Project.iOS.Icons.iPhone_60_3x```.

``` json
 "iOS": {
  "LaunchScreen": {
    "ContentMode": "center",
    "image": "Assets/launch/iOS/Launch.png",
    "width": "200",
    "height": "200"
  }
}
```

> Tip: I noticed when testing on an iOS 9.3 device that the image goes blurry, my only fix was to manually add a copy of the image with a different filename to the Resources folder, then in the LaunchScreen.storyboard, set the Image in the ImageView to the newly added image. iOS13+ works fine.

As of today...

iPhone

77% of all devices introduced in the last four years use iOS 13.
iOS 13: 77%, iOS 12: 17%, Earlier: 6%

70% of all devices use iOS 13.
iOS 13: 70%, iOS 12: 23%, Earlier: 7%

iPad

79% of all devices introduced in the last four years use iPadOS.
iPadOS: 79%, iOS 12: 19%, Earlier: 2%

57% of all devices use iPadOS.
iPadOS: 57%, iOS 12: 27%, Earlier: 16%

Ref: https://developer.apple.com/support/app-store


mortend:
Superseded by #387.

PR #331, opened 2020-05-26
OpenGL clean-up

mortend:
This moves code from `Uno.Runtime.Implementation.ShaderBackends.OpenGL` to `Uno.Graphics.OpenGL` and cleans it up for 2.0.

Fuselibs and other things I have tried still compile fine because these APIs are not used directly.


mortend:
This needs to wait for #329.

PR #338, opened 2020-06-02
Xamarin.Mac v5.14

mortend:
This upgrades Xamarin.Mac from version 3.4 to version 5.14.

PR #341, opened 2020-06-12
mac: replace System.Drawing.Common.dll

mortend:
This references a System.Drawing.Common.dll acquired from the same
source as OpenTK.dll and Xamarin.Mac.dll. This fixes the verbose
runtime-warning introduced in #338.


mortend:
The Travis build has passed: https://travis-ci.org/github/fuse-open/uno/builds/697942771

PR #342, opened 2020-06-13
win: replace OpenTK and ANGLE

mortend:
The OpenTK and ANGLE binaries don't need to update frequently and can be
distributed separately from Uno itself, easily thanks to NPM.

This approach worked well with the Xamarin.Mac binaries I think, and this
will make our future Uno-releases much lighter as well. It's now only 2MB,
and the last version 1.14 release was 16MB.


mortend:
[The Travis build has passed](https://travis-ci.org/github/fuse-open/uno/builds/697944704).

PR #347, opened 2020-06-22
.NET type aliases

mortend:
This will emit assembly attributes containing information about how UnoCore types map to .NET Framework types, when `REFLECTION` is defined. This is helpful to make the implementation of Fuse Simulator more robust.


mortend:
[Travis build succeeded](https://travis-ci.org/github/fuse-open/uno/builds/700931673).

PR #349, opened 2020-06-22
UnoCore: refactor legacy code

mortend:
We want to eliminate legacy C++ libraries eventually in v2.X, and this
is a step in that direction.


mortend:
[Travis build succeeded](https://travis-ci.org/github/fuse-open/uno/builds/701261991).

PR #355, opened 2020-07-09
Mac: update default icon and meta data

mortend:
This uses our Fuse Open logo as the default icon when building macOS apps (`dotnet` and `native`), as we already do on most other build targets.

This also propagate version number and bundle identifier from the project file into macOS apps.


mortend:
[Travis build succeeded](https://travis-ci.org/github/fuse-open/uno/builds/706620287).

Issue #354, opened 2020-07-06
uno.h - Include nested too deeply


ckarmy:
In my app with lastest uno and fuselibs I get this error when I export my app in iOS release:

![image](https://user-images.githubusercontent.com/10090783/86618784-7ef99f00-bf87-11ea-937b-429b98a06b73.png)

My npm:
```
"devDependencies": {
        "@fuse-open/fuselibs": "https://ci.appveyor.com/api/buildjobs/1e5itniolluxteml/artifacts/fuse-open-fuselibs-2.0.0-beta.0.tgz",
        "@fuse-open/uno": "https://ci.appveyor.com/api/buildjobs/18pvechq14p4mbry/artifacts/fuse-open-uno-2.0.0-beta.3.tgz"
    }
```

ckarmy:
Was a problem with git sorry

ckarmy:
I think something related with this commit? https://github.com/fuse-open/uno/commit/52771bace264249ae2319b2c19b8afbd435d8def

mortend:
Does the error persist if you first clean your project before building?

```
uno clean
```

ckarmy:
Yes, I try `uno clean`, clean build folder from Xcode and deleting manualy node_folder, uno folder and build folder. And the error persist...

ckarmy:
I downgrade to this and work fine. But I dont test from which version it fails

```
"devDependencies": {
        "@fuse-open/fuselibs": "https://ci.appveyor.com/api/buildjobs/bubgbup8hs1ytrio/artifacts/fuse-open-fuselibs-2.0.0-beta.0.tgz",
        "@fuse-open/uno": "https://ci.appveyor.com/api/buildjobs/8ife38d4s0i70vpu/artifacts/fuse-open-uno-2.0.0-beta.2.tgz"
},
```

mortend:
Thank you for testing this.

I have made a patched version that I think will solve the problem.

https://ci.appveyor.com/api/buildjobs/fbjn07mn64e39smo/artifacts/fuse-open-uno-2.0.0-beta.3.tgz

ckarmy:
Yes! Now work perfect!!! Thank you very much!!!

PR #356, opened 2020-07-09
UnoCore: use relative include path

mortend:
This fixes a regression #354 introduced in #349.


mortend:
[Travis build succeeded](https://travis-ci.org/github/fuse-open/uno/builds/706417665).

PR #358, opened 2020-07-12
UnoCore: handle exception in Bundle.GetAssembly()

mortend:
This adds some more debug-info if we get problems with assembly loading.


mortend:
[Travis build succeeded](https://travis-ci.org/github/fuse-open/uno/builds/707333188).

PR #361, opened 2020-08-17
Reload config files

mortend:
Reload config files when changed on disk, or every five seconds.

The purpose of the five seconds test is to avoid potential bugs caused by changes in directory structure, i.e. adding or removing config files, that are not being detected by our first test.

The PR fixes a bug in Fuse Studio.


mortend:
https://travis-ci.org/github/fuse-open/uno/builds/718554135

PR #362, opened 2020-08-17
Add BuildTarget.ProjectGroup property 

### LINKED ISSUE #359: unoproj iOS.defines not working in latest Uno beta
Hi guys,

I'm not able to "auto compile" with COCOAPODS flag set in .unoproj via `iOS.defines: ['COCOAPODS']`. So the "defines" probably got disabled somewhere.

mortend:
This returns the prefix that should be used when looking up properties in the project file, for example 'iOS' in 'iOS.Defines'.

Fixes #359.


mortend:
https://travis-ci.org/github/fuse-open/uno/builds/718568701

PR #363, opened 2020-10-02
Android Adaptive Icon

AndrewEQ:
Tutorial: https://medium.com/@andrewq/how-to-use-fuse-open-android-adaptive-icons-6778c41d6b82

Added support for defining an adaptive icon
- still allows legacy definition
- added default fuse open adaptive icon

Notes:
- adaptive icons are only visible in API 26+ (Oreo 8.0.0)
- each icon(foreground/background) must be 108dp*108dp
- easiest solution to get going is to design svg's, then use Android Studio 3's new wizard to generate your image assets: https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive
- in the wizard, you need to name your image asset "icon", then copy over the assets to your fuse project.

.unoproj Usage:

At a minimum, need to set Android.Icons.AdaptiveIconEnabled to true to see the default adaptive icon.

```
"Android": {
    "Icons": {
      "MDPI": "Assets/Icons/Android/mipmap-mdpi/icon.png",
      "HDPI": "Assets/Icons/Android/mipmap-hdpi/icon.png",
      "XHDPI": "Assets/Icons/Android/mipmap-xhdpi/icon.png",
      "XXHDPI": "Assets/Icons/Android/mipmap-xxhdpi/icon.png",
      "XXXHDPI": "Assets/Icons/Android/mipmap-xxxhdpi/icon.png",
      "ANYDPI_v26": "Assets/Icons/Android/mipmap-anydpi-v26/icon.xml",
      "foreground": "Assets/Icons/Android/drawable/icon_foreground.xml",
      "background": "Assets/Icons/Android/drawable/icon_background.xml",

      "MDPI_Round": "AppStoreAssets/Icons/Android/mipmap-mdpi/icon_round.png",
      "HDPI_Round": "AppStoreAssets/Icons/Android/mipmap-hdpi/icon_round.png",
      "XHDPI_Round": "AppStoreAssets/Icons/Android/mipmap-xhdpi/icon_round.png",
      "XXHDPI_Round": "AppStoreAssets/Icons/Android/mipmap-xxhdpi/icon_round.png",
      "XXXHDPI_Round": "AppStoreAssets/Icons/Android/mipmap-xxxhdpi/icon_round.png",
      "ANYDPI_v26_Round": "AppStoreAssets/Icons/Android/mipmap-anydpi-v26/icon_round.xml"
    }
  },
```

To enable the adaptive icon, you must add `-DUSE_ADAPTIVE_ICON` or `-DUSE_ROUND_ADAPTIVE_ICON` to your build command. e.g. `uno build -t=android -DUSE_ADAPTIVE_ICON`


lib/UnoCore/Targets/Android/Android.uxl
```diff
@@ -219,6 +219,38 @@
 
     <!-- Icons -->
 
+    #if @(Project.Android.Icons.AdaptiveIconEnabled:Test(1, 0))
```
mortend, line 222: `#if..#else..#endif` are not supported in UXL. I suggest removing those and the duplicated `<ImageFile>` elements (that are inside the `#else..#endif` block).

lib/UnoCore/Targets/Android/Android.uxl
```diff
@@ -219,6 +219,38 @@
...
+        TargetWidth="48" TargetName="app/src/main/res/mipmap-mdpi/icon.png" />
+    <ImageFile Name="@(Project.Android.Icons.HDPI:Path || '@//Assets/adaptive_icon/mipmap-hdpi/icon.png')"
+        TargetWidth="72" TargetName="app/src/main/res/mipmap-hdpi/icon.png" />
+    <ImageFile Name="@(Project.Android.Icons.XHDPI:Path || '@//Assets/adaptive_icon/mipmap-xhdpi/icon.png')"
+        TargetWidth="96" TargetName="app/src/main/res/mipmap-xhdpi/icon.png" />
+    <ImageFile Name="@(Project.Android.Icons.XXHDPI:Path || '@//Assets/adaptive_icon/mipmap-xxhdpi/icon.png')"
+        TargetWidth="144" TargetName="app/src/main/res/mipmap-xxhdpi/icon.png" />
+    <ImageFile Name="@(Project.Android.Icons.XXXHDPI:Path || '@//Assets/adaptive_icon/mipmap-xxxhdpi/icon.png')"
+        TargetWidth="192" TargetName="app/src/main/res/mipmap-xxxhdpi/icon.png" />
+    <CopyFile Condition="Android" Name="@(Project.Android.Icons.ANYDPI_v26:Path || '@//Assets/adaptive_icon/mipmap-anydpi-v26/icon.xml')" TargetName="app/src/main/res/mipmap-anydpi-v26/icon.xml" />
```
mortend, line 236: `Condition="Android"` is redundant and has no effect. I suggest to remove this attribute.

lib/UnoCore/Targets/Android/Android.uxl
```diff
@@ -219,6 +219,38 @@
...
+        TargetWidth="72" TargetName="app/src/main/res/mipmap-hdpi/icon_round.png" />
+    <ImageFile Name="@(Project.Android.Icons.XHDPI_Round:Path || '@//Assets/adaptive_icon/mipmap-xhdpi/icon_round.png')"
+        TargetWidth="96" TargetName="app/src/main/res/mipmap-xhdpi/icon_round.png" />
+    <ImageFile Name="@(Project.Android.Icons.XXHDPI_Round:Path || '@//Assets/adaptive_icon/mipmap-xxhdpi/icon_round.png')"
+        TargetWidth="144" TargetName="app/src/main/res/mipmap-xxhdpi/icon_round.png" />
+    <ImageFile Name="@(Project.Android.Icons.XXXHDPI_Round:Path || '@//Assets/adaptive_icon/mipmap-xxxhdpi/icon_round.png')"
+        TargetWidth="192" TargetName="app/src/main/res/mipmap-xxxhdpi/icon_round.png" />
+    <CopyFile Condition="Android" Name="@(Project.Android.Icons.ANYDPI_v26_Round:Path || '@//Assets/adaptive_icon/mipmap-anydpi-v26/icon_round.xml')" TargetName="app/src/main/res/mipmap-anydpi-v26/icon_round.xml" />
+
+    <CopyFile Condition="Android" Name="@(Project.Android.Icons.foreground:Path || '@//Assets/adaptive_icon/drawable/icon_foreground.xml')" TargetName="app/src/main/res/drawable/icon_foreground.xml" />
```
mortend, line 250: `Condition="Android"` is redundant and has no effect. I suggest to remove this attribute.

lib/UnoCore/Targets/Android/Android.uxl
```diff
@@ -219,6 +219,38 @@
...
+    <CopyFile Condition="Android" Name="@(Project.Android.Icons.background:Path || '@//Assets/adaptive_icon/drawable/icon_background.xml')" TargetName="app/src/main/res/drawable/icon_background.xml" />
```
mortend, line 251: `Condition="Android"` is redundant and has no effect. I suggest to remove this attribute.

mortend (changes requested):

mortend:
Awesome initiative on making the tutorial. Kudos!

lib/UnoCore/Targets/Android/Android.uxl
AndrewEQ: cool, yeah

lib/UnoCore/Targets/Android/Android.uxl
AndrewEQ: aight...

lib/UnoCore/Targets/Android/Android.uxl
```diff
@@ -221,16 +221,33 @@
 
     <ImageFile Name="@(Project.Android.Icons.LDPI:Path || '@//Assets/DefaultIcon.png')"
         TargetWidth="36" TargetName="app/src/main/res/mipmap-ldpi/icon.png" />
-    <ImageFile Name="@(Project.Android.Icons.MDPI:Path || '@//Assets/DefaultIcon.png')"
+    <ImageFile Name="@(Project.Android.Icons.MDPI:Path || '@//Assets/adaptive_icon/mipmap-mdpi/icon.png')"
```
mortend, line 224: Are we supposed to change the existing default icon from square to round?

From my view it makes sense to keep line 224..232 as they were before, and to delete the `icon.png` variants from the commit.

AndrewEQ: When looking at the history, round is "newer" but looking at the docs again, perhaps I need to create an option for users of round icons.

The android wizard generates the legacy icons too, so I just made use of the new one to keep the look consistent.

AndrewEQ: Ok, I've updated the non-round with the square legacy icons that the wizard generates.

I've also included the option for setting the one round image to then turn on the round icon option: Project.Android.Icons.ANYDPI_v26_Round

AndrewEQ: Hang a sec, I made a typo...

mortend: I mean that the changes on line 224..232 should be reverted thus keeping `DefaultIcon.png` as default icon, like it was before. Then the `icon.png` files in your commit becomes redundant and can be deleted.

AndrewEQ: Oh, ok, can do that, but I just realised a problem...

I added a condition to the manifest that allows a round icon to be defined when a round icon in unoproj property is defined but when trying not to define a round icon, the android.uxl copies the "icon_round.xml" which then tries to look for assets associated with round. I've tried not defining a fuse round default icon to fall back on but the <CopyFile> doesn't accept an empty string. Is there a way to do an if statement/condition in the .uxl so as to not include the round images if not define?

AndrewEQ: Ok, sorted

lib/UnoCore/Targets/Android/Android.uxl
```diff
@@ -231,6 +231,11 @@
         TargetWidth="144" TargetName="app/src/main/res/mipmap-xxhdpi/icon.png" />
     <ImageFile Name="@(Project.Android.Icons.XXXHDPI:Path || '@//Assets/DefaultIcon.png')"
         TargetWidth="192" TargetName="app/src/main/res/mipmap-xxxhdpi/icon.png" />
+    <CopyFile Name="@(Project.Android.Icons.ANYDPI_v26:Path || '@//Assets/adaptive_icon/mipmap-anydpi-v26/icon.xml')" TargetName="app/src/main/res/mipmap-anydpi-v26/icon.xml" />
```
mortend, line 234: I am worried that this would make `icon.xml` unintentionally override user-specified icons in all existing apps, which would be a breaking change. Effectively the change would force all existing apps to provide a new `icon.xml` to restore their original icon.

So after a second though I am not sure if this is the right way forward... Perhaps use `<CopyFile Condition="USE_ADAPTIVE_ICON" ... />`, only copying the new files when `-DUSE_ADAPTIVE_ICON` is specified?

AndrewEQ: Ok, almost there...

AndrewEQ:
Made an option for the round masking: `-DUSE_ROUND_ADAPTIVE_ICON`

mortend (approved):

mortend:
FYI, it's possible provide defines in the project file.

```
"Android": {
    "Defines": [
        "USE_ADAPTIVE_ICON",
        "USE_ROUND_ADAPTIVE_ICON"
    ],
    "Icons": {
        /* ... */
    }
}
```

AndrewEQ:
Oh right, forgot about that, cool will include it in the tutorial too.

PR #368, opened 2020-12-05
Run uno.js in generated build-scripts

mortend:
This will run uno.js instead of '/detected/path/to/mono uno.exe' when
running Uno recursively in generated build scripts (specifically run.sh
for Android).

The reason is that when running from fuse X, '/detected/path/to/mono'
resolves to one of fuse X's executables (because of custom monostub),
and that executable doesn't know how to handle the uno.exe argument,
giving error.


mortend:
https://travis-ci.org/github/fuse-open/uno/builds/747850163

PR #369, opened 2020-12-05
 Fix compile-time warnings on iOS

mortend:
This fixes compile-time warnings seen when building for iOS using latest version of Xcode.


mortend:
https://travis-ci.org/github/fuse-open/uno/builds/747850280

PR #370, opened 2020-12-06
LaunchApk: update reinstall logic

mortend:
Output formatting in 'adb' has changed between versions and there is
no longer a newline before "Failure [...]", breaking our reinstall logic.

Also (IIRC), older versions of 'adb' always exited with code 0 regardless
of success or failure, being the reason we were scanning the output in
the first place.

Now I see that my modern version of 'adb' exits with non-zero code on
failure, so this patch updates the logic to check exit codes instead,
and also logs more details about uninstalling.


mortend:
https://travis-ci.org/github/fuse-open/uno/builds/748084923

PR #371, opened 2020-12-24
Uno.Permission: deprecate all members

mortend:
A fork of this package (Fuse.Android.Permissions) is being merged into
Fuselibs (https://github.com/fuse-open/fuselibs/pull/1387).

Please move to that package instead.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/751370792

PR #372, opened 2020-12-24
iOS: recommend latest version of ios-deploy (not beta)

mortend:
Before it was important to install the beta version because a needed
fix was only included in beta version (related to deploying apps on
a recent version of iOS I believe). This fix is now included in recent
stable versions of ios-deploy so we can safely drop the beta-tag.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/751370828

PR #373, opened 2020-12-24
Build log tweaks

mortend:
Some tweaking to how things are being printed to log during build.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/751373889

PR #375, opened 2020-12-24
UnoCore: new Log class and deprecate assert-statements

mortend:
This adds a new Log class with a cleaner interface to replace the old
Debug class in Uno.Diagnostics.

Also deprecates legacy assert-statements and related functionality in
Uno.Diagnostics.Debug.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/751384645

PR #376, opened 2020-12-24
UnoCore: missing files and iOS warning

mortend:
This adds missing files to the project (after #363) and silences a compile-time warning when building for iOS.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/751384745

PR #378, opened 2020-12-26
unodoc: fix typo in output folder

mortend:
"indicies" is a mistake, so let's fix this typo while at it.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/751546807

PR #380, opened 2020-12-29
Make configuration directory lower-case

mortend:
This makes the configuration part of the build output directory
lower-case, now making all path components of the default build
output directory lower-case.

Some users would argue that all lower-case directories appear more
modern, and look more familiar to users coming from other tooling.
Also, it's not a bad thing to update one's appearance once in a while,
and since we're preparing our next major release this seems like good
timing.

Please note that this is a breaking change (on case-sensitive
file-systems), though unlikely to affect users in practice (since
the majority of our audience are likely users of case-insensitive
file-systems, and it is rarely necessary to deal with your output
directory directly).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,11 +14,11 @@ environment:
 
 install:
   - ps: |
-      if ($env:Target -eq "android") {
+      if ($env:TARGET -eq "android") {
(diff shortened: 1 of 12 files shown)


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/751999432

PR #381, opened 2020-12-29
.NET app-loader clean-ups

mortend:
Some more clean-ups for our .NET app-loaders.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/751999457

PR #383, opened 2021-01-02
UnoCore: forward log to Log class

mortend:
This forwards log-messages from the legacy Debug class to the new
Log class.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/752552132

PR #384, opened 2021-01-14
iOS: fix building for Simulator in Xcode 12

### LINKED ISSUE #364: Unable to run iOS app on Simulator
On XCode 12 later, when building an app, it crashes with a log as seen below:

image

I guess it has something to do with a prebuilt library architecture

mortend:
This fixes building for Simulator in Xcode 12, and also when using
latest Cocoapods v1.17.

* Exclude arm64 to avoid error about missing architectures when using
  Xcode 12.

* Exclude i386 to avoid error about missing architectures when using
  Cocoapods v1.17.

Closes #364


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/754596649

PR #385, opened 2021-01-14
app-win: drop old icon

mortend:
This removes another (and hopefully the last) instance of the old
Fusetools icon.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/754502812

PR #386, opened 2021-01-15
iOS: minor fixes

mortend:
A few patches mainly to silence warnings when working on iOS apps.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/754601463

PR #388, opened 2021-03-09
common: show last line of verbose output in log

mortend:
This will show the last line of verbose output when logging in compact
mode (default), making things slightly more interesting and easier to
see what is happening during time consuming native builds for example.


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/762163671

PR #389, opened 2021-03-23
Update Permissions.uno

AndrewEQ:
Compiles duplicate function names in same class as is already defined in fuselibs/Source/Fuse.Android.Permissions/Permissions.uno:

https://github.com/fuse-open/fuselibs/blob/13f31c90dd69942b722240fc6718de7e37aa7d3d/Source/Fuse.Android.Permissions/Permissions.uno#L108

https://github.com/fuse-open/fuselibs/blob/13f31c90dd69942b722240fc6718de7e37aa7d3d/Source/Fuse.Android.Permissions/Permissions.uno#L115

### DIFF
--- a/src/common/Logging/Log.cs
+++ b/src/common/Logging/Log.cs
@@ -325,6 +325,14 @@ namespace Uno.Logging
         {
             if (level <= Level)
                 WriteLine(line, color);
+            else if (level <= Level + 1)
+            {
+                lock (_state)
+                {
+                    _state.Flush(OutWriter.Inner);
+                    _state.SetLine(OutWriter.Inner, line);
+                }
+            }
         }
 
         public void WriteLine(object line, ConsoleColor? color = null)
@@ -390,6 +398,7 @@ namespace Uno.Logging
             {
                 _state.Flush(writer.Inner);
                 _state.SetColor(writer.Inner, color);
+                _state.SetLine(writer.Inner, null);
                 writer.Inner.WriteLine(line.TrimEnd());
             }
         }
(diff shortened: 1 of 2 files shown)


lib/Uno.Permissions/Permissions.uno
```diff
@@ -110,18 +110,6 @@ namespace Uno.Permissions
             ((IPermissionPromise)promise).Reject(new Exception("Permissions could not be requested or granted."));
         }
 
-        [Foreign(Language.Java), ForeignFixedName]
-        static void permissionRequestSucceeded(object x)
```
mortend, line 114: Suggestion:
```
static void permissionRequestSucceededLegacy(object x)
```

lib/Uno.Permissions/Permissions.uno
```diff
@@ -110,18 +110,6 @@ namespace Uno.Permissions
...
-        @{
-            @{Succeeded(object):Call(x)};
-        @}
-
-        [Foreign(Language.Java), ForeignFixedName]
-        static void permissionRequestFailed(object x)
```
mortend, line 120: Suggestion:
```
static void permissionRequestFailedLegacy(object x)
```

mortend (commented):
Rather than deleting the methods, I think a better solution is to rename them to avoid conflicts. For example by appending `Legacy`.

See also `Permission.java`, lines 97 and 100. Needs to be renamed there as well.

AndrewEQ:
@mortend ok, sorted.

AndrewEQ:
@mortend oh hang on, I didn't check the .java...

AndrewEQ:
@mortend the calls in Permissions.java, wouldn't they end up using the ones defined in fuselibs? Or should I just point them to the legacy functions?

mortend:
> @mortend the calls in Permissions.java, wouldn't they end up using the ones defined in fuselibs? Or should I just point them to the legacy functions?

Point them to the legacy functions please. :)

AndrewEQ:
oops

mortend:
Superseded by #390.

PR #393, opened 2021-04-01
common: improve macOS version detection

mortend:
Uno will now report correct macOS version on Big Sur (11.x).

    Uno version 2.0.0-beta.10 (dev-build) macOS 11.2 x86_64 N/A

Compare with before this patch:

    Uno version 2.0.0-beta.10 (dev-build) macOS 10.16 x86_64 N/A

Also, check ARM architecture (Apple Silicon).


mortend:
- [x] https://travis-ci.org/github/fuse-open/uno/builds/765448836

PR #394, opened 2021-06-11
Android: upgrade build tools and target SDK version

mortend:
Google Play will require this version from August 2021.

https://developer.android.com/distribute/best-practices/develop/target-sdk#pre11


mortend:
- [x] Tested on Fuselibs: https://ci.appveyor.com/project/mortend/fuselibs/builds/39574172

PR #395, opened 2021-08-22
Add ios-sim build target

mortend:
This adds a new build target specifically for the iOS Simulator, so we
can easily build and run apps without going via Xcode.

Example:

    uno build ios-sim --run


mortend:
- [x] https://app.travis-ci.com/github/mortend/uno/builds/235930516

PR #397, opened 2021-08-22
engine: make target defines explicit

mortend:
This is to avoid unintentional defines.


mortend:
- [x] https://app.travis-ci.com/github/mortend/uno/builds/235931498

PR #398, opened 2021-08-22
Add android-emu build target

mortend:
This adds a new build target specifically for emulators (x86_64).

While the regular `android` target builds ARM code by default (not
compatible with emulators), this one builds x86_64 code and makes sure
we run resulting apps on an emulator.

Example:

    uno build android-emu --run

Similar to #395


mortend:
- [x] https://app.travis-ci.com/github/mortend/uno/builds/235954523

PR #399, opened 2021-08-22
Handle C++ exceptions

mortend:
Handle more C++ exceptions to make it easier to spot crashes on startup or in background threads.


mortend:
- [x] https://app.travis-ci.com/github/mortend/uno/builds/235933270

PR #479, opened 2023-05-06
remove xamarin-mac dependency (beta-3.0)

mortend:
Xamarin.Mac isn't supported on .NET 6 (throws exception), blocking us from building iOS apps, so this commit removes the dependency.

Xamarin.Mac was used to access Xcode Developer Teams, so we need a way to replace this functionality. I have searched for a node module or a CLI that can replace it, without luck, so it looks like we'll need to implement a new tool to access Xcode Developer Teams at some point.


mortend:
WIP: https://github.com/mortend/xcode-devteams

PR #508, opened 2023-08-20
Register package visibility needs for google maps

ichan-mb:
On Android targeting API 30 or later, we need to specify in `AndroidManifest` package visibility needs for Google Maps so that in the `MapView` component when we click a marker, the options that will show up on the bottom right of the screen can be clickable and will be redirected to the Google Maps App (currently broken).

This issue has been reported on Slack.

More info about Package Visibility: https://developer.android.com/training/package-visibility/declaring


Anagrr (approved):
Looks good to me

mvondoyannick:
Thanks, let me update

mvondoyannick:
Thanks, let me update

mortend (approved):

